PR #1, opened 2015-07-31
Move preexisting kubernetes plugin and tests

L-MA:


Makefile
```diff
@@ -0,0 +1,45 @@
+.PHONEY: all binary
```
caseydavenport: I think you mean .PHONY

Makefile
```diff
@@ -0,0 +1,45 @@
...
+binary: dist/calico_kubernetes
+test: ut
+
+kubernetesbuild.created: $(BUILD_FILES)
+	cd build_calico_kubernetes; docker build -t calico/kubernetes-build .
+	touch kubernetesbuild.created
+
+dist/calico_kubernetes: kubernetesbuild.created
+	# Build docker container
+	cd build_calico_kubernetes; docker build -t calico/kubernetes-build .
```
caseydavenport: You don't need to build the docker container here if you're depending on kubernetesbuild.created.

Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+BUILD_DIR=build_calico_kubernetes
+BUILD_FILES=$(BUILD_DIR)/Dockerfile $(BUILD_DIR)/requirements.txt
+
+default: all
+all: binary test
+binary: dist/calico_kubernetes
+test: ut
+
+kubernetesbuild.created: $(BUILD_FILES)
```
caseydavenport: Stick a comment here explaining what this does.

Makefile
```diff
@@ -0,0 +1,45 @@
...
+	'/tmp/etcd -data-dir=/tmp/default.etcd/ >/dev/null 2>&1 & \
+	PYTHONPATH=/code/calico_kubernetes nosetests calico_kubernetes/tests -c nose.cfg'
+
+clean:
+	-rm -f *.created
+	find . -name '*.pyc' -exec rm -f {} +
+	-rm -rf dist
+	-docker rm -f calico-build
+	-docker rm -f calico-node
+	-docker rmi calico/node
```
caseydavenport: calico/node shouldn't be in this list.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,353 @@
...
+           bridge
+        3) Create a new calico veth pair, using the docker-assigned IP for the
+           end in the container's namespace
+        4) Assign the node's IP to the host end of the veth pair (required for
+           compatibility with kube-proxy REDIRECT iptables rules).
+        """
+        container_ip = self._read_docker_ip()
+        self._delete_docker_interface()
+        print('Configuring Calico network interface')
+        ep = self.calicoctl('container', 'add', self.docker_id, container_ip, 'eth0')
```
caseydavenport: Don't think this will work the way you expect it to - "calicoctl container add" from the command line is not going to return an Enpoint object like calling container_add( ... ) did.

I'm not sure what the best solution is here, since we want to be replacing this.  Let's chat about this one.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+binary: dist/calico_kubernetes
+test: ut
+
+# Build a new docker image to be used by binary or tests
+kubernetesbuild.created: $(BUILD_FILES)
+	cd build_calico_kubernetes; docker build -t calico/kubernetes-build .
+	touch kubernetesbuild.created
+
+dist/calico_kubernetes: kubernetesbuild.created
+	# Build docker container
```
caseydavenport: Remove the comment here as well - doesn't make sense any more.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	calico/kubernetes-build bash -c \
+	'/tmp/etcd -data-dir=/tmp/default.etcd/ >/dev/null 2>&1 & \
+	PYTHONPATH=/code/calico_kubernetes nosetests calico_kubernetes/tests -c nose.cfg'
+
+clean:
+	-rm -f *.created
+	find . -name '*.pyc' -exec rm -f {} +
+	-rm -rf dist
+	-docker rm -f calico-build
+	-docker rm -f calico-node
```
caseydavenport: Don't need the docker rm -f calico-node line.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,355 @@
...
+
+        try:
+            addr = addrs[0]
+            print('Using IP Address %s' % (addr))
+            return addr
+        except IndexError:
+            # If both get_host_ips return empty lists, print message and exit.
+            print('No Valid IP Address Found for Host - cannot configure networking for pod %s' % (self.pod_name))
+            sys.exit(1)
+
```
caseydavenport: Remove the extra newline here while you're at it.

caseydavenport:
A couple more comments to address, but pretty close.

caseydavenport:
LGTM

PR #2, opened 2015-07-31
Remove calls to calicoctl except container calls

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -326,14 +346,18 @@ def _apply_tags(self, profile_name, pod):
...
+
         for k, v in labels.iteritems():
             tag = '%s_%s' % (k, v)
             tag = tag.replace('/', '_')
             print('Adding tag ' + tag)
-            try:
-                self.calicoctl('profile', profile_name, 'tag', 'add', tag)
-            except sh.ErrorReturnCode as e:
-                print('Could not create tag %s.\n%s' % (tag, e))
+            self._datastore_client.profile_update_tags(profile)
```
caseydavenport: You're missing the line that actually adds the tag to the profile.

Right now, this just gets the profile, then saves the profile :)

calico_kubernetes/calico_kubernetes.py
```diff
@@ -53,12 +55,47 @@ def delete(self, pod_name, docker_id):
         self.pod_name = pod_name
         self.docker_id = docker_id
 
+        if not self._datastore_client.profile_exists(self.pod_name):
+            print "A profile with name %s does not exist" % self.pod_name
+            sys.exit(1)
```
caseydavenport: We don't want to sys.exit here.

I think we should be trying to delete the profile, catching the exception if it doesn't exist, logging out, and then continuing.

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -174,8 +178,8 @@ def test_delete_docker_interface(self):
             m_check_output.assert_has_calls(calls)
 
     def test_configure_profile(self):
-        with patch.object(self.plugin, 'calicoctl',
-                    autospec=True) as m_calicoctl, \
+        with patch.object(self.plugin, '_datastore_client',
```
caseydavenport: One thing you could do to avoid repeating the patch of the datastore client is patch it once when you create self.plugin.  There isn't ever a case where we want to actually call into the datastore client, so this would be fine.

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -16,8 +16,10 @@
 import unittest
 from mock import patch, Mock, call
 from subprocess import CalledProcessError
+import docker
```
caseydavenport: Don't think you need to import docker here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -300,11 +322,16 @@ def _apply_rules(self, profile_name):
         :type profile_name: string
         :return:
         """
+        try:
+            profile = self._datastore_client.get_profile(profile_name)
+        except:
+            print "Could not apply rules. Profile not found."
```
caseydavenport: Log out the name of the profile in this error message - it will be more useful that way.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -2,12 +2,16 @@
...
 import sys
+import docker
+from docker import Client
 from subprocess import check_output, CalledProcessError, check_call
 import requests
 import sh
 
 # Append to existing env, to avoid losing PATH etc.
 # Need to edit the path here since calicoctl loads client on import.
+DOCKER_VERSION = "1.16"
```
caseydavenport: I don't think this is the right place for this, just because the comment above doesn't apply to this line.

Could you move `DOCKER_VERSION=...` away from that comment?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -58,7 +67,41 @@ def delete(self, pod_name, docker_id):
...
+        """
+        Configure the calico profile for a pod.
+
+        Currently assumes one pod with each name.
+        """
+        print('Configuring Pod Profile')
+        profile_name = self.pod_name
+
+        if self._datastore_client.profile_exists(profile_name):
+            print "A profile with name %s already exists" % profile_name
```
caseydavenport: Anywhere that you sys.exit, I think that needs to be clear in the logs.

So, instead of "A profile with name already exists", maybe "ERROR: A profile with name already exists, exiting."  Or something similar.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -326,16 +353,33 @@ def _apply_tags(self, profile_name, pod):
...
         print('Finished applying tags.')
 
+    def _get_container_id(self, container_name):
+        try:
+            info = self._docker_client.inspect_container(container_name)
+        except docker.errors.APIError as e:
+            if e.response.status_code == 404:
+                print "Container %s was not found." % container_name
+            else:
+                print e.message
```
caseydavenport: "ERROR: Could not get container ID - %s" % e.message

Might be a more clear error.

calico_kubernetes/tests/kube_plugin_test.py
caseydavenport: Don't think you need to import Profile either - you don't need to import things that are being Mocked out.

caseydavenport:
LGTM

PR #3, opened 2015-07-31
Calico Policy in Kubernetes Annotations

alexhersh:
move PR from old repo (https://github.com/Metaswitch/calico-docker/pull/401)

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -6,11 +6,12 @@ from docker import Client
 from docker.errors import APIError
 from subprocess import check_output, CalledProcessError, check_call
 import requests
+from urllib import quote
 import sh
+import re
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.datastore_datatypes import Rules
-
 
 DOCKER_VERSION = "1.16"
 
@@ -62,7 +63,7 @@ class NetworkPlugin(object):
         self.pod_name = pod_name
         self.docker_id = docker_id
 
-        print('Deleting container %s with profile %s' % 
+        print('Deleting container %s with profile %s' %
             (self.docker_id, self.pod_name))
 
         # Remove the profile for the workload.
@@ -91,7 +92,7 @@ class NetworkPlugin(object):
 
         pod = self._get_pod_config()
 
-        self._apply_rules(profile_name)
+        self._apply_rules(profile_name, pod)
 
         self._apply_tags(profile_name, pod)
 
@@ -142,7 +143,7 @@ class NetworkPlugin(object):
         return ep
 
     def _get_node_ip(self):
-        """ 
+        """
         Determine the IP for the host node.
         """
         # Compile list of addresses on network, return the first entry.
@@ -269,31 +270,64 @@ class NetworkPlugin(object):
         auth_data = json.loads(json_string)
         return auth_data['BearerToken']
 
-    def _generate_rules(self):
+    def _generate_rules(self, pod):
         """
-        Generate the Profile rules that have been specified on the Pod's ports.
+        Generate Rules takes human readable policy strings in annotations
+        and creates argument arrays for calicoctl
 
-        We only create a Rule for a port if it has 'allowFrom' specified.
-
-        The Rule is structured to match the Calico etcd format.
-
-        :return list() rules: the rules to be added to the Profile.
+        :return two lists of rules(arg lists): inbound list of rules (arg lists)
+        outbound list of rules (arg lists)
         """
+
+        namespace, ns_tag = self._get_namespace_and_tag(pod)
+
+        # kube-system services need to be accessed by all namespaces
+        if namespace == "kube-system" :
+            print "using kube-system, allow all"
+            return [["allow"]], [["allow"]]
+
         inbound_rules = [
-            {
-                'action': 'allow',
-            },
+            ["allow", "from", "tag", ns_tag]
         ]
 
         outbound_rules = [
-            {
-                'action': 'allow',
-            },
+            ["allow"]
         ]
+
+        print("Getting Policy Rules from Annotation of pod %s" % pod)
+
+        annotations = self._get_metadata(pod, "annotations")
+
+        # Find policy block of annotations
+        if annotations and "policy" in annotations.keys():
+            # Remove Default Rule (Allow Namespace)
+            inbound_rules = []
+            rules = annotations["policy"]
+
+            # Rules separated by semicolons
+            for rule in rules.split(";"):
+                args = rule.split(" ")
+
+                # Replace labels with tags
+                # key=value -> namespace_key_value
+                # TODO: account for improper formatting
+                if 'label' in args:
+                    label_ind = args.index('label')
+                    args[label_ind] = 'tag'
+                    label = args[label_ind + 1]
+                    key, value = label.split('=')
+                    tag = self._label_to_tag(key, value, namespace)
+                    args[label_ind + 1] = tag
+
+                # Remove null args and add to rule list
+                args = filter(None, args)
+                inbound_rules.append(args)
+
         return inbound_rules, outbound_rules
 
     def _generate_profile_json(self, profile_name, rules):
         """
+        DEPRECIATED: Not used in current semantic annotation format
         Given a list of of Calico rules, generate a Calico Profile JSON blob
         implementing those rules.
 
@@ -311,10 +345,10 @@ class NetworkPlugin(object):
             'outbound_rules': outbound,
         }
         profile_json = json.dumps(profile, indent=2)
-        print('Final profile "%s": %s' % (profile_name, profile_json))
+        print('Final profile "%s":\n%s' % (profile_name, profile_json))
         return profile_json
 
-    def _apply_rules(self, profile_name):
+    def _apply_rules(self, profile_name, pod):
         """
         Generate a new profile with the default 'allow all' rules.
 
@@ -328,10 +362,33 @@ class NetworkPlugin(object):
             print("Error: Could not apply rules. Profile not found: %s, exiting" % profile_name)
             sys.exit(1)
 
-        rules = self._generate_rules()
-        profile_json = self._generate_profile_json(profile_name, rules)
-        profile.rules = Rules.from_json(profile_json)
-        self._datastore_client.profile_update_rules(profile)
+        inbound_rules, outbound_rules = self._generate_rules(pod)
+
+        print "Removing Default Rules"
+
+        # TODO: This method is append-only, not profile replacement, we need to remove default rules
+        # TODO: remove calicoctl calls (x5)
+        try:
+            self.calicoctl('profile', profile_name, 'rule', 'remove', 'inbound', '--at=2')
+            self.calicoctl('profile', profile_name, 'rule', 'remove', 'inbound', '--at=1')
+            self.calicoctl('profile', profile_name, 'rule', 'remove', 'outbound', '--at=1')
+        except sh.ErrorReturnCode as e:
+            print('Could not delete default rules for profile %s (assumed 2 inbound, 1 outbound)\n%s' % (profile_name, e))
+
+        for rule in inbound_rules:
+            print 'applying inbound rule \n%s' % rule
+            try:
+                self.calicoctl('profile', profile_name, 'rule', 'add', 'inbound', rule)
+            except sh.ErrorReturnCode as e:
+                print('Could not create rule %s.\n%s' % (rule, e))
+
+        for rule in outbound_rules:
+            print 'applying outbound rule \n%s' % rule
+            try:
+                self.calicoctl('profile', profile_name, 'rule', 'add', 'outbound', rule)
+            except sh.ErrorReturnCode as e:
+                print('Could not create rule %s.\n%s' % (rule, e))
+
         print('Finished applying rules.')
 
     def _apply_tags(self, profile_name, pod):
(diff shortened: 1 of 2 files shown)


alexhersh:
TODO: handle pod info more efficiently

calico_kubernetes/calico_kubernetes.py
```diff
@@ -257,17 +258,42 @@ def _generate_rules(self):
...
+                "action": "allow"
+            }
         ]
+
+        print('Getting Policy Rules from Annotation of pod %s' % pod)
+
+        annotations = self._get_metadata(pod, 'annotations')
+
+        if 'allowFrom' in annotations.keys():
+            inbound_rules = []
```
caseydavenport: Stick a comment here about why we're blatting over the other inbound_rules, so we don't forget.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -257,17 +258,44 @@ def _generate_rules(self):
...
+        if 'allowFrom' in annotations.keys():
+            # Remove Default Rule (Allow Namespace)
+            inbound_rules = []
+            rules = json.loads(annotations['allowFrom'])
+            for rule in rules:
+                rule['action'] = 'allow'
+                rule = self._translate_rule(rule, namespace)
+                inbound_rules.append(rule)
+
+        if 'allowTo' in annotations.keys():
```
caseydavenport: Don't implement the 'allowTo' bits right now, since we haven't defined them in the design.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -319,23 +347,120 @@ def _apply_tags(self, profile_name, pod):
...
+        """
+        try:
+            val = pod['metadata'][key]
+            print("%s of pod %s:\n%s" % (key, pod, val))
+            return val
+        except KeyError:
+            print('No %s found in pod %s' % (key, pod))
+            return None
+
+    def _escape_chars(self, tag):
```
caseydavenport: nitpicky - this is supposed to be generic function, so instead of naming the argument 'tag', maybe make it 'string' or 'unescaped'

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,31 +270,64 @@ def _get_api_token(self):
...
+        :return two lists of rules(arg lists): inbound list of rules (arg lists)
+        outbound list of rules (arg lists)
+        """
 
-        The Rule is structured to match the Calico etcd format.
+        namespace, ns_tag = self._get_namespace_and_tag(pod)
+
+        # kube-system services need to be accessed by all namespaces
+        if namespace == "kube-system" :
+            print "using kube-system, allow all"
```
caseydavenport: "Pod %s belongs to the kube-system namespace - allow all inbound and outbound traffic"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,31 +270,64 @@ def _get_api_token(self):
...
+            inbound_rules = []
+            rules = annotations["policy"]
+
+            # Rules separated by semicolons
+            for rule in rules.split(";"):
+                args = rule.split(" ")
+
+                # Replace labels with tags
+                # key=value -> namespace_key_value
+                # TODO: account for improper formatting
```
caseydavenport: What is this TODO referring to?

In general, we don't want TODOs left around in the code.  If this is something you think needs to be done for this PR, lets talk.  Otherwise, stick in on the backlog somewhere we can remember it and we'll get to it later.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,31 +270,64 @@ def _get_api_token(self):
...
+                if 'label' in args:
```
caseydavenport: I'd stick a more helpful comment (or series of comments) here explaining what the following code is doing.

```
# Labels are declared in the annotations with the format 'label X=Y' - these must be converted into format 'tag NAMSPACE_X_Y' to be parsed by calicoctl.
...
# Replace "label" with "tag"
...
# Split the provided label into component bits
...
# Form a Calico tag from the given key:value pair
...
```

The code isn't that complex, but even so nicely commented code is a lot easier to follow, looks prettier, and will be that much easier to refactor in a month when we've forgotten all about this.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,31 +270,64 @@ def _get_api_token(self):
...
+                    label_ind = args.index('label')
+                    args[label_ind] = 'tag'
+                    label = args[label_ind + 1]
+                    key, value = label.split('=')
+                    tag = self._label_to_tag(key, value, namespace)
+                    args[label_ind + 1] = tag
+
+                # Remove null args and add to rule list
```
caseydavenport: What "null args" are you referring to here?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,31 +270,64 @@ def _get_api_token(self):
...
+                args = filter(None, args)
+                inbound_rules.append(args)
+
         return inbound_rules, outbound_rules
 
     def _generate_profile_json(self, profile_name, rules):
         """
+        DEPRECIATED: Not used in current semantic annotation format
```
caseydavenport: If it isn't used, lets just get rid of it.  It will still exist in our git history, so we can always go back to it.  Prevents clutter and the confusion of "why is this still here?"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -328,10 +362,33 @@ def _apply_rules(self, profile_name):
...
-        profile_json = self._generate_profile_json(profile_name, rules)
-        profile.rules = Rules.from_json(profile_json)
-        self._datastore_client.profile_update_rules(profile)
+        inbound_rules, outbound_rules = self._generate_rules(pod)
+
+        print "Removing Default Rules"
+
+        # TODO: This method is append-only, not profile replacement, we need to remove default rules
+        # TODO: remove calicoctl calls (x5)
+        try:
```
caseydavenport: This is a bit weird.

We should know exactly what rules are created by default, correct?

If you want to be resilient to changes, you might want to do this in a while loop, and break when you hit the specific error thrown when there are no more rules.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -311,10 +345,10 @@ def _generate_profile_json(self, profile_name, rules):
...
         }
         profile_json = json.dumps(profile, indent=2)
-        print('Final profile "%s": %s' % (profile_name, profile_json))
+        print('Final profile "%s":\n%s' % (profile_name, profile_json))
         return profile_json
 
-    def _apply_rules(self, profile_name):
+    def _apply_rules(self, profile_name, pod):
         """
         Generate a new profile with the default 'allow all' rules.
```
caseydavenport: Is this really what this method does?

Looks more like it takes an existing profile, removes the default rules, and then applies the correct inbound and outbound rules as generated in _generate_rules.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -328,10 +362,33 @@ def _apply_rules(self, profile_name):
...
+            self.calicoctl('profile', profile_name, 'rule', 'remove', 'outbound', '--at=1')
+        except sh.ErrorReturnCode as e:
+            print('Could not delete default rules for profile %s (assumed 2 inbound, 1 outbound)\n%s' % (profile_name, e))
+
+        for rule in inbound_rules:
+            print 'applying inbound rule \n%s' % rule
+            try:
+                self.calicoctl('profile', profile_name, 'rule', 'add', 'inbound', rule)
+            except sh.ErrorReturnCode as e:
+                print('Could not create rule %s.\n%s' % (rule, e))
```
caseydavenport: Could not apply __inbound__ rule

caseydavenport: I also like to preface these kind of things with "ERROR" so that they are really apparent in the logs.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -346,12 +403,6 @@ def _apply_tags(self, profile_name, pod):
         :return:
```
caseydavenport: Worth putting some more into the docstring here.  It should describe what the default set of tags are, and what the generated tags will be (maybe even with an example!)

calico_kubernetes/calico_kubernetes.py
```diff
@@ -359,16 +410,72 @@ def _apply_tags(self, profile_name, pod):
...
-        for k, v in labels.iteritems():
-            tag = '%s_%s' % (k, v)
-            tag = tag.replace('/', '_')
-            print('Adding tag ' + tag)
-            profile.tags.add(tag)
-
+        # Grab namespace and create a tag if it exists.
+        namespace, ns_tag = self._get_namespace_and_tag(pod)
+
+        if ns_tag:
```
caseydavenport: I think we should have some sort of ERROR or WARNING here if ns_tag is None - that should never really happen as far as I know.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -359,16 +410,72 @@ def _apply_tags(self, profile_name, pod):
...
+        print("%s of pod %s:\n%s" % (key, pod, val))
+        return val
+
+    def _escape_chars(self, unescaped_string):
+        """
+        Calico can only handle 3 special chars, '_.-'
+        This function uses regex sub to replace SCs with '_'
+        """
+        escape_seq = '_'
+        return re.sub('[^a-zA-Z0-9\.-]', escape_seq, unescaped_string)
```
caseydavenport: You mention that Calico supports _, ., and -, yet you still replace those characters in the substitution?

I also think we should be differentiating _ from and escaped _.  e.g. stick an extra underscore before any underscores in the string.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -359,16 +410,72 @@ def _apply_tags(self, profile_name, pod):
...
+        namespace = self._get_metadata(pod, 'namespace')
+        ns_tag = self._escape_chars('%s=%s' % ('namespace', namespace)) if namespace else None
+        return namespace, ns_tag
+
+    def _label_to_tag(self, label_key, label_value, namespace):
+        """
+        Labels are key-value pairs, tags are single strings. This function handles that translation
+        1) concatenate key and value with '='
+        2) prepend a pod's namespace followed by '/' if available
+        3) replace special characters with urllib-style escape sequence
```
caseydavenport: Third bullet point here is wrong - can just say "escape the generated string so it is Calico compatible" -- the implementation of the escape logic is irrelevant.

calico_kubernetes/calico_kubernetes.py
alexhersh: kubernetes will fill in namespace=default if no namespace is specified. 
Also if namespace is blank for some reason, _get_metadata will issue an warning and return None, which is handled by _get_ns_and_tag

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,75 +270,127 @@ def _get_api_token(self):
...
+        inbound_rules, outbound_rules = self._generate_rules(pod)
+
+        print "Removing Default Rules"
+
+        # TODO: This method is append-only, not profile replacement, we need to replace calicoctl calls
+        #       but necessary functions are not available in pycalico ATM
+
+        # Remove default rules. Assumes 2 in, 1 out.
+        try:
+            self.calicoctl('profile', profile_name, 'rule', 'remove', 'inbound', '--at=2')
```
caseydavenport: Not a big fan of this approach.  

if we know there are always three, then we can do something like loop three times, and remove the first rule each iteration.

caseydavenport: Still pretty gross, since what if there are 4?  We should probably iterate until there are no rules left.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,75 +270,127 @@ def _get_api_token(self):
...
 
-        rules = self._generate_rules()
-        profile_json = self._generate_profile_json(profile_name, rules)
-        profile.rules = Rules.from_json(profile_json)
-        self._datastore_client.profile_update_rules(profile)
+        inbound_rules, outbound_rules = self._generate_rules(pod)
+
+        print "Removing Default Rules"
+
+        # TODO: This method is append-only, not profile replacement, we need to replace calicoctl calls
```
caseydavenport: If you do the bit below, then it would replace the profile (albeit in a slow hacky way) and we could remove this comment.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,75 +272,127 @@ def _get_api_token(self):
...
+
+        # kube-system services need to be accessed by all namespaces
+        if namespace == "kube-system" :
+            print "Pod %s belongs to the kube-system namespace - allow all inbound and outbound traffic" % (pod)
+            return [["allow"]], [["allow"]]
+
+        if namespace:
+            inbound_rules = [["allow", "from", "tag", ns_tag]]
+            outbound_rules = [["allow"]]
+        else:
```
caseydavenport: When are we expecting namespace to be None?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -269,75 +272,127 @@ def _get_api_token(self):
...
-            },
-        ]
         return inbound_rules, outbound_rules
 
-    def _generate_profile_json(self, profile_name, rules):
+    def _apply_rules(self, profile_name, pod):
         """
-        Given a list of of Calico rules, generate a Calico Profile JSON blob
-        implementing those rules.
+        Generate a rules for a given profile based on annotations.
```
caseydavenport: typo - "Generate a rules"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -359,16 +408,81 @@ def _apply_tags(self, profile_name, pod):
...
+        return val
+
+    def _escape_chars(self, unescaped_string):
+        """
+        Calico can only handle 3 special chars, '_.-'
+        This function uses regex sub to replace SCs with '_'
+        """
+        # Character to replace symbols
+        swap_char = '_'
+        # If swap_char is in string, double it.
```
caseydavenport: Nitpicky formatting thing - all in-line comments should follow a newline.

PR #6, opened 2015-08-05
Replace calicoctl container methods

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+        :param interface: The name of the interface in the container.
+        """
+        # Obtain information from Docker Client and validate container state
+        info = self._get_container_info(container_name)
+        self._validate_container_add(info)
+
+        # Obtain ip pool from datastore client, validate pool, and assign
+        # ip address through IPAM Client
+        ip = IPAddress(ip)
+        self._assign_address_in_pool_IPAM(ip)
```
djosborne: In the mesos plugin we assign multiple IPv4 addresses, and optional IPv6 addresses in the pool. Does Kuberentes allow for that? Regardless, for commonized functionality, might want to make `_assign_address_in_pool_IPAM` accept an array of IP addresses

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
         print('Finished configuring network interface')
         return ep
 
+    def _configure_profile(self, endpoint):
```
caseydavenport: You've got this defined twice, in two different places, and you've removed it from its original location.  Could you clean this up so that its back in the original location?

it will make the diff smaller, and thus easier to review.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -326,16 +576,33 @@ def _apply_tags(self, profile_name, pod):
...
-                self.calicoctl('profile', profile_name, 'tag', 'add', tag)
-            except sh.ErrorReturnCode as e:
-                print('Could not create tag %s.\n%s' % (tag, e))
+            profile.tags.add(tag)
+
+        self._datastore_client.profile_update_tags(profile)
+
         print('Finished applying tags.')
 
+    def _get_container_id(self, container_name):
```
caseydavenport: You've got this defined in two places as well.

caseydavenport:
Could you fix up the conflicts before I do any more reviewing?  Looks like this PR has a bunch of code thats mixed up with your last PR, which makes it looks pretty weird.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+        else:
+            ep.ipv6_nets.add(network)
+            ep.ipv6_gateway = next_hop
+
+        return ep
+
+    def _endpoint_complete(self, ep, pid, ip, interface):
+        # Create the veth, move into the container namespace, add the IP and
+        # set up the default routes.
+        pid = info["State"]["Pid"]
```
caseydavenport: Info is declared here - I think you actually want to remove this line, and just use the give pid argument.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+                      workload_id=container_id,
+                      endpoint_id=uuid.uuid1().hex,
+                      state="active",
+                      mac=None)
+        if network.version == 4:
+            ep.ipv4_nets.add(network)
+            ep.ipv4_gateway = next_hop
+        else:
+            ep.ipv6_nets.add(network)
+            ep.ipv6_gateway = next_hop
```
caseydavenport: I think you need to set the endpoint at the end of this method as well.  Not required for the Kubernetes plugin, but it is likely required for other plugins (such as Mesos) which don't call endpoint_complete directly after endpoint_init.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+        return ep
+
+    def _container_remove(self, container_name):
+        """
+        Remove the indicated container on this host from Calico networking
+
+        :param container_name: THe name or partial id of the container
+        """
+        # Resolve the name to ID.
+        workload_id = self._get_container_id(container_name)
```
caseydavenport: Don't call _get_container_id(...) inside `container_add` or `container_remove`.  This relies on the docker_client, and we want to make sure that these APIs are not reliant on docker.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+    def _container_add(self, container_name, ip, interface):
+        """
+        Add a container (on this host) to Calico networking with the given IP.
+
+        :param container_name: The name or ID of the container.
+        :param ip: An IPAddress object with the desired IP to assign.
+        :param interface: The name of the interface in the container.
+        """
+        # Obtain information from Docker Client and validate container state
+        info = self._get_container_info(container_name)
```
caseydavenport: Same with `_get_container_info` - this relies on docker, so shouldn't be part of the `container_add` method.

djosborne: Can we move the `get_container_info` call out of `_container_add`, and instead add parameters to the function for `container_id` and `pid`? For mesos, those parameters aren't available through the docker client

djosborne: Oop,  my comment is a well timed duplicate of Casey's.

calico_kubernetes/calico_kubernetes.py
djosborne: `datastore.set_endpoint(ep)`

calico_kubernetes/calico_kubernetes.py
caseydavenport: Ignore this bit - we were wrong.  `container_add` is not intended to go into pycalico, so this is fine.

djosborne:
This PR only supports adding a single IP IPv4 address to an endpoint. Just noting that we'll need to add Multi-IPv4 support as well as IPv6 support later (probably a PR I will open with the Mesos work)

djosborne:
I don't think this needs to be a NetworkPlugin class. Since the plugin binary calls are one-shots, state isn't saved between calls. I think this should either be moved to have global variables and functions, or at least make the methods `@static` to instead of `NetworkPlugin().create(pod_name, docker_id)` we have `NetworkPlugin.create(pod_name, docker_id)`

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+        if not info["State"]["Running"]:
+            print("The container is not currently running.")
+            sys.exit(1)
+
+        # We can't set up Calico if the container shares the host namespace.
+        if info["HostConfig"]["NetworkMode"] == "host":
+            print("Can't add the container to Calico because it is running NetworkMode = host.")
+            sys.exit(1)
+
+    def _assign_address_in_pool_IPAM(self, ip):
```
caseydavenport: Just a note on names - I think `_assign_address(...)` and `_free_address(...)` are more suitable.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+
+    def _unassign_address_in_pool_IPAM(self, ip):
+        pools = self._datastore_client.get_ip_pools(ip.version)
+        for pool in pools:
+            if ip in pool:
+                # Ignore failure to unassign address, since we're not
+                # enforcing assignments strictly in datastore.py.
+                self._ipam_client.unassign_address(pool, ip)
+
+    def _endpoint_init(self, container_id, ip):
```
caseydavenport: This should probably be called `create_endpoint`, and then `endpoint_complete` can be renamed to `create_veth`.  This lines up more closely with what they're actually doing.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+
+        # Remove the endpoint
+        netns.remove_veth(endpoint.name)
+
+        # Remove the container from the datastore.
+        self._datastore_client.remove_workload(hostname, ORCHESTRATOR_ID, workload_id)
+
+        print("Removed Calico interface from %s" % container_name)
+
+    def _validate_container_add(self, info):
```
caseydavenport: I might structure this differently.

I think the validation code is not something we want to stick in pycalico - this is just a nice to have, and might vary per-orchestrator.


I'd rename this to `_validate_container_info(self, info)` which validates that the container info is good.  The bit of validation that checks that the endpoint doesn't exist could probably just be inlined in _container_add.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -97,6 +147,223 @@ def _configure_interface(self):
...
+         # The next hop IPs for this host are stored in etcd.
+        next_hops = self._datastore_client.get_default_next_hops(hostname)
+        try:
+            next_hop = next_hops[ip.version]
+        except KeyError:
+            print("This node is not configured for IPv%d." % ip.version)
+            sys.exit(1)
+
+        network = IPNetwork(IPAddress(ip))
+        ep = Endpoint(hostname=hostname,
```
caseydavenport: hostname and ORCHESTRATOR_ID are only available in this context because they are globals in the calico_kubernetes file.  If this gets moved to pycalico, we'll need to pass in the ORCHESTRATOR_ID and the hostname as arguments.

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: all binary ut clean
+.PHONY: all binary ut clean dist/calico_kubernetes
```
caseydavenport: A better way of doing this is probably to just replace all instances of `dist/calico_kubernetes` with `binary`.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +128,233 @@ def _configure_interface(self):
...
+                hostname=hostname,
+                orchestrator_id=orchestrator_id,
+                workload_id=workload_id
+            )
+        except KeyError:
+            # Calico doesn't know about this container.  Continue.
+            pass
+        else:
+            # Calico already set up networking for this container.  Since we got
+            # called with an IP address, we shouldn't just silently exit, since
```
caseydavenport: I don't think this big block comment is really applicable here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +128,233 @@ def _configure_interface(self):
...
+
+        # Remove the endpoint
+        netns.remove_veth(endpoint.name)
+
+        # Remove the container from the datastore.
+        self._datastore_client.remove_workload(hostname, orchestrator_id, workload_id)
+
+        print("Removed Calico interface from %s" % container_name)
+
+    def _validate_container_add(self, container_name):
```
caseydavenport: A better name for this method is probably something like `_validate_container_state` - it isn't specifically tied to the container_add logic at all.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +128,233 @@ def _configure_interface(self):
...
+            info = self._docker_client.inspect_container(container_name)
+        except APIError as e:
+            if e.response.status_code == 404:
+                print("Container %s was not found." % container_name)
+            else:
+                print(e.message)
+            sys.exit(1)
+        return info
+
+    def _get_container_id(self, container_name):
```
caseydavenport: This method has a whole bunch of duplication in it!

What about this?

```
def _get_container_id(self, container_name):
  return self._get_container_info(container_name)["Id"]
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +128,233 @@ def _configure_interface(self):
...
+        self._datastore_client.set_endpoint(ep)
+
+        return ep
+
+    def _get_container_info(self, container_name):
+        try:
+            info = self._docker_client.inspect_container(container_name)
+        except APIError as e:
+            if e.response.status_code == 404:
+                print("Container %s was not found." % container_name)
```
caseydavenport: Remember to make it clear when we're about to exit - could lead to confusing diagnostics otherwise.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +128,233 @@ def _configure_interface(self):
...
+            else:
+                print(e.message)
+            sys.exit(1)
+        return info["Id"]
+
+    def _get_container_pid(self, container_name):
```
caseydavenport: Same thing here, regarding duplication.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Still not 100% sure why this has moved - could you stick _configure_profile back up at line 85 where it belongs? :)

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +158,119 @@ def _configure_interface(self):
...
+
+    def _get_container_id(self, container_name):
+        return self._get_container_info(container_name)["Id"]
+
+    def _get_container_pid(self, container_name):
+        return self._get_container_info(container_name)["State"]["Pid"]
+
+    def _read_docker_ip(self):
+        """Get the IP for the pod's infra container."""
+        ip = self._get_container_info(self.docker_id)["NetworkSetting"]["IPAddress"]
```
caseydavenport: Nice catch here - thats some feel-good commonization if I've ever seen it.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +158,119 @@ def _configure_interface(self):
...
+
+        # Obtain ip pool from datastore client, validate pool, and assign
+        # ip address through IPAM Client
+        self._ipam_client.assign_address(None, ip)
+
+        # Obtain next hops from datastore client, validate next hop, and
+        # create Endpoint object
+        ep = self._datastore_client.create_endpoint(hostname, orchestrator_id,
+                                                    workload_id, ip)
+        self._datastore_client.set_endpoint(ep)
```
caseydavenport: I don't think you need this set_endpoint() here.

If you replace the update_endpoint below with a set_endpoint, we should be fine.  No need to do two database writes instead of one.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,8 +158,119 @@ def _configure_interface(self):
         print('Finished configuring network interface')
         return ep
 
+    def _container_add(self, container_name, workload_id, pid, ip, interface,
```
caseydavenport: There methods were called "container_add" and "container_remove" in calicoctl / dockerland.  However, for kubernetes, I think we want to call these something pod-centric.

e.g. `add_pod(...)` and `remove_pod(...)`

caseydavenport: I think we can trim down these arguments a lot.

caseydavenport:
Needs a rebase

calico_kubernetes/calico_kubernetes.py
```diff
@@ -37,7 +44,9 @@ def __init__(self):
         self.profile_name = None
         self.namespace = None
         self.docker_id = None
+
         self._datastore_client = DatastoreClient()
+        self._ipam_client = IPAMClient()
```
djosborne: IPAMClient inherits from DatastoreClient. So any calls you make through datastore_client can just be made through ipam_client. (right @caseydavenport ?)

caseydavenport: Ya got that right. 

`self.client`

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +165,117 @@ def _configure_interface(self):
...
+                       hostname, orchestrator_id):
+        """
+        Add a container (on this host) to Calico networking with the given IP.
+
+        :param container_name: The name or ID of the container.
+        :param ip: An IPAddress object with the desired IP to assign.
+        :param interface: The name of the interface in the container.
+        """
+        # Check if the container already exists. If it does, exit.
+        try:
```
djosborne: A cleaner, more readable way to do this is to use get_endpoints, which will return an empty list instead of raising an error.
```
if len(self._datastore_client.get_endpoint(
                hostname=hostname,
                orchestrator_id=orchestrator_id,
                workload_id=workload_id)) == 1:
      print("This container has already been configured with Calico Networking.")
      sys.exit(1)
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +165,117 @@ def _configure_interface(self):
...
+        return ep
+
+    def _container_remove(self, container_name, workload_id, hostname, orchestrator_id):
+        """
+        Remove the indicated container on this host from Calico networking
+
+        :param container_name: THe name or partial id of the container
+        """
+        # Find the endpoint ID. We need this to find any ACL rules
+        try:
```
djosborne: You can leave this try/except. It makes sense here

calico_kubernetes/calico_kubernetes.py
caseydavenport: Eh, not sure that makes a significant different.  It saves a few lines perhaps, but the exception handling is fine I think.

build_calico_kubernetes/requirements.txt
```diff
@@ -1,6 +1,7 @@
 PyInstaller==2.1
-git+https://github.com/projectcalico/libcalico.git
+git+https://github.com/projectcalico/libcalico.git@master
```
caseydavenport: We don't want to be tracking master here in our dependencies - we should track a specific version of libcalico that we know works.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +165,117 @@ def _configure_interface(self):
...
+        # Obtain information from Docker Client and validate container state
+        self._validate_container_state(container_name)
+
+        # Obtain ip pool from datastore client, validate pool, and assign
+        # ip address through IPAM Client
+        self._ipam_client.assign_address(None, ip)
+
+        # Obtain next hops from datastore client, validate next hop, and
+        # create Endpoint object
+        ep = self._datastore_client.create_endpoint(hostname, orchestrator_id,
```
caseydavenport: We should be at least somewhat resilient to errors here.  If any of these calls fail, we'll have allocated an IP address and will never release it.

These should catch any expected errors and release the assigned IP address if they fail.

calico_kubernetes/calico_kubernetes.py
djosborne: I'd keep `self.datstore_client`. We have a `docker_client` and `etcd_client` elsewhere, don't want to get confused

Makefile
```diff
@@ -47,6 +47,7 @@ ut-circle: binary
 	--with-xunit --xunit-file=/circle_output/output.xml; RC=$$?;\
 	[[ ! -z "$$COVERALLS_REPO_TOKEN" ]] && coveralls || true; exit $$RC'
 
+
```
caseydavenport: Extra newline

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+        else:
+            print("This container has already been configured with Calico Networking.")
+            sys.exit(1)
+
+        # Obtain information from Docker Client and validate container state
+        self._validate_container_state(container_name)
+
+        # Assign ip address through IPAM Client
+        try:
+            _ = self._datastore_client.assign_address(None, ip)
```
caseydavenport: If you're using the the value returned by a function / method, you should give it a name.  Only use underscores to indicate that the value is never accessed.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+        except PoolNotFound:
+            print "ERROR: IP address %s does not belong to any configured pools."\
+                  " Exiting." % ip
+            sys.exit(1)
+        else:
+            if _ is False:
```
caseydavenport: No need to check if a boolean value is False.

Should do this:

```
if not success:
  ...
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+        try:
+            ep = self._datastore_client.create_endpoint(hostname, orchestrator_id,
+                                                        workload_id, [ip])
+        except AddrFormatError:
+            print "This node is not configured for IPv%d. Unassigning IP "\
+                  "address %s then exiting."  % ip.version, ip
+            self._datastore_client.unassign_address(None, ip)
+            sys.exit(1)
+
+        self._datastore_client.set_endpoint(ep)
```
caseydavenport: Don't need to set the endpoint at this intermediate stage - you can just set it once at the end of this processing.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+        # Assign ip address through IPAM Client
+        try:
+            _ = self._datastore_client.assign_address(None, ip)
+        except PoolNotFound:
+            print "ERROR: IP address %s does not belong to any configured pools."\
+                  " Exiting." % ip
+            sys.exit(1)
+        else:
+            if _ is False:
+               print "ERROR: Failed to assign IP address %s. Exiting." % ip
```
caseydavenport: You don't actually exit here, though you say you're going to :)

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+            print "This node is not configured for IPv%d. Unassigning IP "\
+                  "address %s then exiting."  % ip.version, ip
+            self._datastore_client.unassign_address(None, ip)
+            sys.exit(1)
+
+        self._datastore_client.set_endpoint(ep)
+
+        # Create the veth, move into the container namespace, add the IP and
+        # set up the default routes.
+        # Update Endpoint in felix
```
caseydavenport: Remove the "Update Endpoint in Felix bit"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+        ep.mac = ep.provision_veth(pid, interface)
+        self._datastore_client.update_endpoint(ep)
+        # Let the caller know what endpoint was created.
```
caseydavenport: Stick a newline before any comments.

We shouldn't ever have code that is directly followed by a comment.

```
code_here()

# Then a comment
some_more_code(...)

# Another comment
and_code()
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +162,130 @@ def _configure_interface(self):
...
+                  "address %s then exiting."  % ip.version, ip
+            self._datastore_client.unassign_address(None, ip)
+            sys.exit(1)
+
+        self._datastore_client.set_endpoint(ep)
+
+        # Create the veth, move into the container namespace, add the IP and
+        # set up the default routes.
+        # Update Endpoint in felix
+        ep.mac = ep.provision_veth(pid, interface)
```
caseydavenport: What happens if this fails?  I think we need to:

1) Not update the endpoint
2) unassign the address.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -67,13 +73,14 @@ def delete(self, namespace, pod_name, docker_id):
         self.pod_name = pod_name
         self.docker_id = docker_id
         self.namespace = namespace
+        workload_id = self._get_container_id(self.docker_id)
```
caseydavenport: I'm pretty sure we don't need to call _get_container_id here - self.docker_id should already be the container ID, as far as I know.

caseydavenport: A number of other places as well.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -122,12 +129,20 @@ def _configure_interface(self):
         4) Assign the node's IP to the host end of the veth pair (required for
            compatibility with kube-proxy REDIRECT iptables rules).
         """
+        # Set up parameters
+        container_name = self.docker_id
+        workload_id = self._get_container_id(container_name)
+        container_pid = self._get_container_pid(container_name)
         container_ip = self._read_docker_ip()
-        container_id = self._get_container_id(self.docker_id)
+        container_ip = IPAddress(container_ip)
```
caseydavenport: _read_docker_ip() should return an IPAddress object so that you don't need to wrap it in one here.

calico_kubernetes/calico_kubernetes.py
L-MA: provision_veth does not raise any exceptions so how would we know if it failed?

caseydavenport: Hm, maybe not something to deal with then.  We should have some way of knowing if this operation succeeded or failed though.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +157,118 @@ def _configure_interface(self):
...
+
+        # Assign ip address through IPAM Client
+        try:
+            ip_assigned = self._datastore_client.assign_address(None, ip)
+        except PoolNotFound:
+            print "ERROR: IP address %s does not belong to any configured pools."\
+                  " Exiting." % ip
+            sys.exit(1)
+        else:
+            if ip_assigned is False:
```
caseydavenport: This is a bit weird - the pythonic way to write this is going to be

```
if not ip_assigned:
  ...
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +157,118 @@ def _configure_interface(self):
...
+               print "ERROR: Failed to assign IP address %s. Exiting." % ip
```
caseydavenport: Still missing an exit here..

calico_kubernetes/calico_kubernetes.py
```diff
@@ -147,6 +157,118 @@ def _configure_interface(self):
...
+                workload_id=self.docker_id
+            )
+        except KeyError:
+            print("Container %s doesn't contain any endpoints" % self.docker_id)
+            sys.exit(1)
+
+        # Remove any IP address assignments that this endpoint has
+        for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
+            assert(net.size == 1)
+            ip = net.ip
```
caseydavenport: No need to create a whole new variable here - combine this line with the one below it.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -2,18 +2,23 @@
...
+from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
-from pycalico.datastore_datatypes import Rules
+from pycalico.ipam import IPAMClient
+from pycalico.datastore_errors import PoolNotFound
 
 DOCKER_VERSION = "1.16"
+ORCHESTRATOR_ID = "docker"
+hostname = socket.gethostname()
```
caseydavenport: hostname --> HOSTNAME

All caps makes it more clear that this is a global.

Alternatively, make it "self.hostname" and add it to NetworkPlugin.__init__

caseydavenport:
LGTM.

PR #7, opened 2015-08-05
Elements needed for CircleCI

L-MA:


Makefile
```diff
@@ -32,6 +35,20 @@ ut: dist/calico_kubernetes
 	'/tmp/etcd -data-dir=/tmp/default.etcd/ >/dev/null 2>&1 & \
 	PYTHONPATH=/code/calico_kubernetes nosetests calico_kubernetes/tests -c nose.cfg'
 
+# UT runs on Cicle need to create the calicoctl binary
```
caseydavenport: I don't think UT runs on Circle need to create the calicoctl binary for the calico_kubernetes repo.

Makefile
```diff
@@ -32,6 +35,20 @@ ut: dist/calico_kubernetes
...
+ut-circle: kubernetesbuild.created dist/calico_kubernetes
+	# Can't use --rm on circle
+	# Circle also requires extra options for reporting.
+	docker run \
+	-v `pwd`/calico_kubernetes:/code/calico_kubernetes \
+	-v $(CIRCLE_TEST_REPORTS):/circle_output \
+	-e COVERALLS_REPO_TOKEN=$(COVERALLS_REPO_TOKEN) \
+	calico/kubernetes-build bash -c \
+	'/tmp/etcd -data-dir=/tmp/default.etcd/ >/dev/null 2>&1 & \
```
caseydavenport: We don't need this bit about running etcd - the kubernetes plugin tests don't rely on etcd being present / running.

Makefile
```diff
@@ -25,12 +27,26 @@ dist/calico_kubernetes: kubernetesbuild.created
 	-e PYTHONPATH=/code/calico_kubernetes \
 	calico/kubernetes-build pyinstaller calico_kubernetes/calico_kubernetes.py -a -F -s --clean
 
-ut: dist/calico_kubernetes
+ut: binary
```
caseydavenport: Don't think the UTs rely on the binary being built.

Makefile
```diff
@@ -25,12 +27,26 @@ dist/calico_kubernetes: kubernetesbuild.created
...
 	docker run --rm -v `pwd`/calico_kubernetes:/code/calico_kubernetes \
 	-v `pwd`/nose.cfg:/code/nose.cfg \
 	calico/kubernetes-build bash -c \
-	'/tmp/etcd -data-dir=/tmp/default.etcd/ >/dev/null 2>&1 & \
-	PYTHONPATH=/code/calico_kubernetes nosetests calico_kubernetes/tests -c nose.cfg'
+	'>/dev/null 2>&1 & PYTHONPATH=/code/calico_kubernetes \
+	nosetests calico_kubernetes/tests -c nose.cfg'
+
+# UT runs on Cicle
+ut-circle: binary
```
caseydavenport: Fine to make the ut-circle rely on it though, so that we can assert the binary gets built when Circle is checking stuff.

caseydavenport: It will need to rely on the kubernetesbuild.created job though.

caseydavenport:
Lookin' good to me.  A couple more comments on the makefile, but then good to go.

Makefile
```diff
@@ -1,11 +1,13 @@
 .PHONY: all binary ut clean
 
+SRCDIR=calico_kubernetes
+CALICO_KUBERNETES=$(wildcard $(SRCDIR)/*.py)
```
caseydavenport: Whats up with these guys? What is the wildcard line needed for?

PR #18, opened 2015-08-18
Replace print statements with logging

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -172,7 +194,7 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
             # Calico doesn't know about this container.  Continue.
             pass
         else:
-            print("This container has already been configured with Calico Networking.")
+            logger.info("This container has already been configured with Calico Networking.")
```
caseydavenport: Anywhere that we sys.exit() should probably be error level logging.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -254,9 +276,9 @@ def _get_container_info(self, container_name):
             info = self._docker_client.inspect_container(container_name)
         except APIError as e:
             if e.response.status_code == 404:
-                print("Error: Container %s was not found. Exiting." % container_name)
+                logger.errror("Container %s was not found. Exiting." % container_name)
```
caseydavenport: When using python logging, you can ditch the "%" operator to do string formatting.

Its more efficient to use the following syntax, because then the string formatting only occurs if the log is emitted (for example, debug logs won't get emitted when logging is set to INFO)

```
logger.error("Container %s was not found", container_name)
```

caseydavenport: Also, you've got a typo: logger.errror

calico_kubernetes/calico_kubernetes.py
```diff
@@ -591,20 +614,37 @@ def _label_to_tag(self, label_key, label_value):
         tag = self._escape_chars(tag)
         return tag
 
+
+def config_logger():
+    logging._srcfile = None
+    log_dir = '/var/log/calico/kubernetes'
+    if not os.path.exists(log_dir):
+        os.mkdir(log_dir)
```
caseydavenport: use `os.makedirs()` here instead - this will only work if /var/log/calico already exists.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -591,20 +614,37 @@ def _label_to_tag(self, label_key, label_value):
         tag = self._escape_chars(tag)
         return tag
 
+
+def config_logger():
+    logging._srcfile = None
+    log_dir = '/var/log/calico/kubernetes'
```
caseydavenport: Define this guy as a constant at the top,  rather than locally here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -591,20 +614,37 @@ def _label_to_tag(self, label_key, label_value):
         tag = self._escape_chars(tag)
         return tag
 
+
+def config_logger():
```
caseydavenport: I might actually move this to a new file, "logging.py" or something, to keep it separated from plugin code.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -9,13 +9,18 @@
...
+import logging
 from netaddr import IPAddress, AddrFormatError
+from calico_logging import config_logger
 from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
 from pycalico.datastore_errors import PoolNotFound
 
+LOG_DIR = '/var/log/calico/kubernetes'
```
caseydavenport: I'd move LOG_DIR to calico_logging.  It doesn't need to be configurable from this file.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -9,13 +9,18 @@
 from subprocess import check_output, CalledProcessError, check_call
 import requests
 import sh
+import logging
 from netaddr import IPAddress, AddrFormatError
+from calico_logging import config_logger
```
caseydavenport: rename calico_logging to logutils

calico_kubernetes/calico_kubernetes.py
```diff
@@ -85,7 +90,25 @@ def delete(self, namespace, pod_name, docker_id):
         try:
             self._datastore_client.remove_profile(self.profile_name)
         except:
-            print "Cannot remove profile %s; Profile cannot be found." % self.profile_name
+            logger.warning("Cannot remove profile %s; Profile cannot be found.",
+                           self.profile_name)
+
+    def status(self, namespace, pod_name, docker_id):
```
caseydavenport: As discussed before, could you keep the logging / status hook / IPAM code in separate pull requests?

The logging shouldn't be dependent on the status code being right.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -58,14 +63,14 @@ def create(self, namespace, pod_name, docker_id):
...
+        logger.info('Configuring docker container %s', self.docker_id)
 
         try:
             endpoint = self._configure_interface()
             self._configure_profile(endpoint)
         except CalledProcessError as e:
-            print('Error code %d creating pod networking: %s\n%s' % (
-                e.returncode, e.output, e))
+            logger.error('Error code %d creating pod networking: %s\n%s',
+                e.returncode, e.output, e)
```
caseydavenport: Could you fix up the indentation here?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -591,20 +616,25 @@ def _label_to_tag(self, label_key, label_value):
         tag = self._escape_chars(tag)
         return tag
 
+
 if __name__ == '__main__':
-    print('Args: %s' % sys.argv)
+    config_logger(logger, LOG_DIR)
```
caseydavenport: Move this guy to the top right after you create the logger.

calico_kubernetes/calico_logging.py
```diff
@@ -0,0 +1,13 @@
+#!/bin/python
+import logging
+import os
+
+def config_logger(logger, log_dir):
```
caseydavenport: nitpick: we can spell out configure, it looks nicer.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -95,10 +97,11 @@ def _configure_profile(self, endpoint):
         """
         pod = self._get_pod_config()
 
-        print('Configuring Pod Profile: %s' % self.profile_name)
+        logger.info('Configuring Pod Profile: %s' % self.profile_name)
```
caseydavenport: still using a % operator here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -9,13 +9,17 @@
...
+import logging
 from netaddr import IPAddress, AddrFormatError
+from logutils import configure_logger
 from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
 from pycalico.datastore_errors import PoolNotFound
 
+logger = logging.getLogger('kubernetes')
```
caseydavenport: don't name the logger 'kubernetes' - maybe calico_kubernetes, or just use __name__

caseydavenport:
A couple more comments, then LGTM.

calico_kubernetes/logutils.py
```diff
@@ -0,0 +1,18 @@
...
+import logging
+import os
+
+LOG_DIR = '/var/log/calico/kubernetes'
+
+def configure_logger(logger, log_dir=None):
+    if log_dir is None:
+        log_dir = LOG_DIR
+
+    logging._srcfile = None
```
caseydavenport: I'm not familiar with this _srcfile bit.  What does it do?

L-MA: Its not really necessary but by setting _srcfile to None, the logger does not collect information on where the calls are coming from

PR #20, opened 2015-08-19
Create status method that prints ip info for kubernetes

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -89,6 +89,24 @@ def delete(self, namespace, pod_name, docker_id):
...
+        # Obtain IP Address of container
+        ip = self._read_docker_ip()
+
+        json_dict = {
+            "apiVersion": "v1beta1",
+            "kind": "PodNetworkStatus",
+            "ip": str(ip)
+        }
+
+        logger.info("Writing json dict to stdout: \n%s", json.dumps(json_dict))
```
caseydavenport: This log is more useful if it says its writing the pod status to stdout.

So something like "Writing status for pod %s: %s"

caseydavenport:
LGTM - not much point in merging this on its own, but its good to do it in steps.

PR #23, opened 2015-08-21
Add Calico IPAM to plugin

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -14,9 +14,12 @@
 from logutils import configure_logger
 from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
+from pycalico.datastore_errors import DataStoreError
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
-from pycalico.datastore_errors import PoolNotFound
+from pycalico.netns import PidNamespace
+
+logger = logging.getLogger(__name__)
```
caseydavenport: Duplicate lines here

calico_kubernetes/calico_kubernetes.py
```diff
@@ -35,6 +38,11 @@
 KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
                                'http://kubernetes-master:8080/api/v1/')
 
+# Flag to indicate whether or not to use Calico IPAM
+# If False, use the default docker container ip address to create container
+# If True, use libcalico's auto_assign IPAM to create container
+CALICO_IPAM = os.environ.get('CALICO_IPAM', 'default')
```
caseydavenport: I think this should just be a boolean value.

CALICO_IPAM="true" or CALICO_IPAM="false - it should default to "false"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -89,6 +97,45 @@ def delete(self, namespace, pod_name, docker_id):
...
+            endpoint = self._datastore_client.get_endpoint(
+                hostname=HOSTNAME,
+                orchestrator_id=ORCHESTRATOR_ID,
+                workload_id=self.docker_id
+            )
+        except KeyError:
+            logger.error("Container %s doesn't contain any endpoints" % self.docker_id)
+            sys.exit(1)
+
+        # Remove any IP address assignments that this endpoint has
```
caseydavenport: I think this comment is wrong.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -89,6 +97,45 @@ def delete(self, namespace, pod_name, docker_id):
...
+        logger.info("Retrieved IP Address: %s", ip)
+
+        json_dict = {
+            "apiVersion": "v1beta1",
+            "kind": "PodNetworkStatus",
+            "ip": str(ip)
+        }
+
+        logger.info("Writing json dict to stdout: \n%s", json.dumps(json_dict))
+
```
caseydavenport: Don't need the newline after the logger line.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,32 +227,43 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
         self._validate_container_state(self.docker_id)
 
-        # Assign ip address through IPAM Client
-        try:
-            ip_assigned = self._datastore_client.assign_address(None, ip)
-        except PoolNotFound:
-            logger.error("IP address %s does not belong to any configured "
-                         "pools. Exiting.", ip)
-            sys.exit(1)
+        if CALICO_IPAM == 'calico':
```
caseydavenport: You could simplify container_add by moving the IP address assignment into its own method.

```
def _get_container_ip(self):
   if CALICO:
    <use IPAM>
  else:
   read_docker_ip()
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,32 +227,43 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+            logger.info("Using Calico IPAM")
```
caseydavenport: nitpick: logging should go after the comment.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,32 +227,43 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+            # Assign ip address through IPAM Client
+            try:
+                ip_lists = self._datastore_client.auto_assign_ips(
```
caseydavenport: instead of assigning the return value to a single variable, you can break it into two vars inline.

```
ipv4_addrs, ipv6_addrs = self._datastore_client.auto_assign_ips( ... )
assert not ipv6_addrs  # We didn't ask for any IPv6 addresses
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,32 +227,43 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+                1, 0, str(self.docker_id), None)
+            except RuntimeError as err:
+                logger.error("Cannot auto assign IPAddress: %s", err.message)
+
+            # Concatenate lists into one single list
+            ip_list = ip_lists[0]+ip_lists[1]
```
caseydavenport: I don't think we need to concat these - we don't care about the IPv6 addresses for now, so lets just throw them away.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,32 +227,43 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
-            self._datastore_client.unassign_address(None, ip)
+                         "IP address %s then exiting.", ip_list[0].version, ip_list[0])
+            self._datastore_client.release_ips(set(ip_list))
             sys.exit(1)
 
         # Create the veth, move into the container namespace, add the IP and
         # set up the default routes.
-        ns = netns.PidNamespace(pid)
-        ep.mac = ep.provision_veth(ns, interface)
+        logger.info("Creating the veth with pid %s on interface %s", pid, interface)
```
caseydavenport: nitpick - the veth doesn't have a PID, it is the namespace's PID.

So, this log should be "Creating veth in namespace pid=%s with interface name %s" % (pid, interface)

calico_kubernetes/calico_kubernetes.py
caseydavenport: This whole bit could be commonized into a method called "assign_container_ip" which is called by container add - simplifies the container_add code a bit.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -228,15 +285,30 @@ def _container_remove(self, hostname, orchestrator_id):
...
         # Remove any IP address assignments that this endpoint has
+        ip_set = set()
         for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
             assert(net.size == 1)
-            self._datastore_client.unassign_address(None, net.ip)
+            ip_set.add(net.ip)
 
-        # Remove the endpoint
-        netns.remove_veth(endpoint.name)
+        logger.info("Removing IP address %s from endpoint %s", ip_set, endpoint.name)
```
caseydavenport: typo: address -> addresses

calico_kubernetes/calico_kubernetes.py
```diff
@@ -14,9 +14,11 @@
 from logutils import configure_logger
 from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
+from pycalico.datastore_errors import DataStoreError
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
-from pycalico.datastore_errors import PoolNotFound
+
+logger = logging.getLogger(__name__)
```
caseydavenport: You do logging.getLogger() twice here.

caseydavenport: Could you change "logger" to "_log" ?

It will standardize the naming with our other repos.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,37 +225,64 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+                assert not ipv6_addrs
+            except RuntimeError as err:
+                logger.error("Cannot auto assign IPAddress: %s", err.message)
+                sys.exit(1)
+        else:
+            logger.info("Using docker assigned IP address")
+            ip_list = [self._read_docker_ip()]
+            try:
+                self._datastore_client.assign_ip(ip_list[0], str(self.docker_id), None)
+            except Exception:
```
caseydavenport: Don't catch all exceptions here. In general, its bad to do "Pokemon style" exception handling (gotta catch 'em all), since it can hide bugs.  Just catch what you expect to be thrown.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -228,15 +299,30 @@ def _container_remove(self, hostname, orchestrator_id):
...
 
-        # Remove the endpoint
-        netns.remove_veth(endpoint.name)
+        logger.info("Removing IP addresses %s from endpoint %s", ip_set, endpoint.name)
+        self._datastore_client.release_ips(ip_set)
 
-        # Remove the container from the datastore.
-        self._datastore_client.remove_workload(hostname, orchestrator_id, self.docker_id)
+        # Remove the veth interface from endpoint
+        if netns.remove_veth(endpoint.name):
```
caseydavenport: remove_veth(...) doesn't return False when it hits an error - it raises a CalledProcessError.  It will return false if the veth doesn't exist.  From our perspective, we don't care about this case (since we wan't the veth to not exist anyway).

Also - In general, it's usually not a good idea to add extra branches just for logging.

I might do the following instead:
```
_log.info("Removing veth interface from endpoint %s" % endpoint.name)
try:
  netns.remove_veth(endpoint.name)
except:
  _log.exception(...)
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -228,15 +299,30 @@ def _container_remove(self, hostname, orchestrator_id):
...
+            logger.warning("Could not remove veth interface from endpoint %s",
+                           endpoint.name)
+
+        # Remove the container/endpoint from the datastore.
+        try:
+            self._datastore_client.remove_workload(
+                hostname, orchestrator_id, self.docker_id)
+            logger.info("Successfully removed workload from datastore")
+        except KeyError as err:
+            logger.warning("Could not remove workload from endpoint %s",
```
caseydavenport: This log is a bit weird - we're not removing the workload from the endpoint, we're just removing the workload.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -228,15 +299,30 @@ def _container_remove(self, hostname, orchestrator_id):
...
+            logger.info("Successfully removed veth interface from endpoint %s",
+                        endpoint.name)
+        else:
+            logger.warning("Could not remove veth interface from endpoint %s",
+                           endpoint.name)
+
+        # Remove the container/endpoint from the datastore.
+        try:
+            self._datastore_client.remove_workload(
+                hostname, orchestrator_id, self.docker_id)
```
caseydavenport: We've got a weird mix of passing in `hostname` vs `HOSTNAME` (same for orchestrator_id).

I'd opt for removing some of the arguments to our functions, and just always use the globally defined version when calling into libcalico.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -181,37 +225,64 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+        if CALICO_IPAM == 'true':
+            # Assign ip address through IPAM Client
+            logger.info("Using Calico IPAM")
+            try:
+                ip_list, ipv6_addrs = self._datastore_client.auto_assign_ips(
+                    1, 0, None, None)
+                logger.info("ip_list is %s; ipv6_addrs is %s", ip_list, ipv6_addrs)
+                assert not ipv6_addrs
+            except RuntimeError as err:
+                logger.error("Cannot auto assign IPAddress: %s", err.message)
```
caseydavenport: These should both be `_log.exception()`

calico_kubernetes/calico_kubernetes.py
```diff
@@ -604,18 +690,26 @@ def _label_to_tag(self, label_key, label_value):
...
+            if mode == 'setup':
+                logger.info('Executing Calico pod-creation hook')
+                NetworkPlugin().create(namespace, pod_name, docker_id)
+            elif mode == 'teardown':
+                logger.info('Executing Calico pod-deletion hook')
+                NetworkPlugin().delete(namespace, pod_name, docker_id)
+            elif mode == 'status':
+                logger.info('Executing Calico pod-status hook')
+                NetworkPlugin().status(namespace, pod_name, docker_id)
+    except:
```
caseydavenport: I know we talked about this - in general it is bad to catch all exceptions.  In this case, it is OK, but we should add a comment explaining why we're doing this.

This is an "exception" to that rule, since we're re-raising the exception right after - we're just capturing it so that we can get the log outpupt.

build_calico_kubernetes/requirements.txt
```diff
@@ -1,4 +1,6 @@
 PyInstaller==2.1
+git+https://github.com/projectcalico/libcalico.git
+git+https://github.com/projectcalico/python-etcd.git
```
caseydavenport: Do we use this requirement?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -40,6 +42,11 @@
 # Allow the user to enable/disable namespace isolation policy
 DEFAULT_POLICY = os.environ.get('DEFAULT_POLICY', 'allow')
 
+# Flag to indicate whether or not to use Calico IPAM
```
caseydavenport: Make sure to use proper punctuation on your comments! That means periods at the end!

calico_kubernetes/calico_kubernetes.py
```diff
@@ -94,6 +101,44 @@ def delete(self, namespace, pod_name, docker_id):
...
+                hostname=HOSTNAME,
+                orchestrator_id=ORCHESTRATOR_ID,
+                workload_id=self.docker_id
+            )
+        except KeyError:
+            logger.error("Container %s doesn't contain any endpoints" % self.docker_id)
+            sys.exit(1)
+
+        # Retrieve IPAddress from the attached IPNetworks on the endpoint
+        # Since Kubernetes only support ipv4, we'll only check for ipv4 nets
```
caseydavenport: typo: support -> supports

calico_kubernetes/calico_kubernetes.py
```diff
@@ -186,62 +228,104 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
 
         # Create Endpoint object
         try:
-            ep = self._datastore_client.create_endpoint(hostname, orchestrator_id,
-                                                        self.docker_id, [ip])
-        except AddrFormatError:
-            logger.error("This node is not configured for IPv%d. Unassigning "
-                         "IP address %s then exiting.", ip.version, ip)
-            self._datastore_client.unassign_address(None, ip)
+            logger.info("Creating endpoint for ips in list %s", ip_list)
```
caseydavenport: "... with IPs %s" rather than "for ips"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -186,62 +228,104 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+            ep = self._datastore_client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
+                                                        self.docker_id, ip_list)
+        except (AddrFormatError, KeyError):
+            logger.exception("Failed to create endpoint in list %s. Unassigning "
```
caseydavenport: I'd tweak to: "Failed to create endpoint with IPs %s. Unassigning"

calico_kubernetes/calico_kubernetes.py
```diff
@@ -186,62 +228,104 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+        else:
+            logger.info("Using docker assigned IP address")
+            ip_list = [self._read_docker_ip()]
+            try:
+                self._datastore_client.assign_ip(ip_list[0], str(self.docker_id), None)
+            except (ValueError, RuntimeError, AlreadyAssignedError):
+                logger.exception("Cannot assign IPAddress %s", ip_list[0])
+                sys.exit(1)
+
+        return ip_list
```
caseydavenport: We never expect this to return more than 1 IP address.  I'd change the signature of this method to return a single IP.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -186,62 +228,104 @@ def _container_add(self, pid, ip, interface, hostname, orchestrator_id):
...
+        ip_set = set()
         for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
             assert(net.size == 1)
-            self._datastore_client.unassign_address(None, net.ip)
+            ip_set.add(net.ip)
 
-        # Remove the endpoint
-        netns.remove_veth(endpoint.name)
+        logger.info("Removing IP addresses %s from endpoint %s", ip_set, endpoint.name)
+        self._datastore_client.release_ips(ip_set)
```
caseydavenport: You could clean this up quite a bit by removing the assert(net.size==1) above (which doesn't buy us much).

```
# Release all IPs on this endpoint.
ip_set = set(endpoint.ipv4_nets | endpoint.ipv6_nets)
self._datastore_client.release_ips(ip_set)
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -619,9 +701,19 @@ def _label_to_tag(self, label_key, label_value):
         namespace = sys.argv[2].replace('/', '_')
         pod_name = sys.argv[3].replace('/', '_')
         docker_id = sys.argv[4]
+
+        logger.info('Args: %s' % sys.argv)
+        logger.debug("Using ETCD_AUTHORITY=%s", os.environ[ETCD_AUTHORITY_ENV])
```
caseydavenport: I think these are worth keeping at info level.

calico_kubernetes/logutils.py
```diff
@@ -1,17 +1,30 @@
...
 import os
+import pycalico
+
+_log = logging.getLogger(pycalico.__name__)
 
 LOG_DIR = '/var/log/calico/kubernetes'
+ROOT_LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(message)s'
+LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(filename)s: %(message)s'
+
+def configure_logger(logger, root_logger=False, log_dir=None):
```
caseydavenport: Could you add a docstring that describes how you expect this to be used, and explains the arguments and their defaults?

caseydavenport: It's fine to set `log_dir=LOG_DIR` as a default here, since its just a string.

build_calico_kubernetes/requirements.txt
```diff
@@ -1,4 +1,5 @@
 PyInstaller==2.1
+git+https://github.com/projectcalico/libcalico.git
```
caseydavenport: Could you pin this to the latest release, rather than master?

caseydavenport:
One more comment, then I think we're good.

calico_kubernetes/logutils.py
```diff
@@ -3,15 +3,48 @@
...
+        formatter = logging.Formatter(ROOT_LOG_FORMAT)
+        hdlr.setFormatter(formatter)
+    else:
+        formatter = logging.Formatter(LOG_FORMAT)
+        hdlr.setFormatter(formatter)
+
     logger.addHandler(hdlr)
-    logger.setLevel(logging.INFO)
+
+    if logging_level == 'notset':
```
caseydavenport: I think a nicer way to do this would be to pass in the correct logging.XXXXX constant in the `log_level` parameter so you can get rid of the giant if/else block.

In the calling code, you can define the mapping from user-input to log-level like so:

```
log_levels = { "DEBUG": logging.DEBUG,
  "INFO": logging.INFO,
  ...
}
```

Then, you can do things like:
```
log_level = log_levels.get(LOG_LEVEL, logging.INFO)
```

Symmetric: I don't think you need to maintain this mapping. You can pass strings into logging.setLevel; see logging/__init__.py:_levelNames for the valid strings (they just need to be uppercased I think).

caseydavenport: Ah, even better.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -12,13 +12,21 @@
...
 from pycalico import netns
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
-from pycalico.datastore_errors import PoolNotFound
+from pycalico.block import AlreadyAssignedError
 
 logger = logging.getLogger(__name__)
+pycalico_logger = logging.getLogger(pycalico.__name__)
+log_levels = {"DEBUG": logging.DEBUG,
```
caseydavenport: As @Symmetric said in his comment, you don't actually need the mapping I suggested, since setLevel accepts a string.

You can simply do:
```
logging_level = os.environ.get("LOG_LEVEL", "INFO")
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -604,12 +689,9 @@ def _label_to_tag(self, label_key, label_value):
 
 
 if __name__ == '__main__':
-    configure_logger(logger)
-    logger.info('Args: %s' % sys.argv)
-    logger.info("Using ETCD_AUTHORITY=%s", os.environ[ETCD_AUTHORITY_ENV])
-    logger.info("Using CALICOCTL_PATH=%s", CALICOCTL_PATH)
-    logger.info("Using KUBE_API_ROOT=%s", KUBE_API_ROOT)
-    logger.info("Using DEFAULT_POLICY=%s", DEFAULT_POLICY)
+    configure_logger(logger, LOG_LEVEL.upper(), True)
```
caseydavenport: You could simplify this by moving the .upper() call above to where you are reading the environment variable.

So,
```
LOG_LEVEL = env.get(...).upper()
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -619,9 +701,20 @@ def _label_to_tag(self, label_key, label_value):
         namespace = sys.argv[2].replace('/', '_')
         pod_name = sys.argv[3].replace('/', '_')
         docker_id = sys.argv[4]
+
+        logger.info('Args: %s' % sys.argv)
+        logger.info("Using ETCD_AUTHORITY=%s", os.environ[ETCD_AUTHORITY_ENV])
+        logger.info("Using CALICOCTL_PATH=%s", CALICOCTL_PATH)
+        logger.info("Using KUBE_API_ROOT=%s", KUBE_API_ROOT)
+        logger.info("Using CALICO_IPAM=%s", CALICO_IPAM)
+        logger.info("Using DEFAULT_POLICY=%s", DEFAULT_POLICY)
```
caseydavenport: You should log out the logging level here as well.

PR #24, opened 2015-08-24
Env variable to disable/enable namespace isolation policy

L-MA:


calico_kubernetes/calico_kubernetes.py
```diff
@@ -35,6 +35,9 @@
 KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
                                'http://kubernetes-master:8080/api/v1/')
 
+# Allow the user to enable/disable namespace isolation policy
+NAMESPACE_ISOLATION = os.environ.get('NAMESPACE_ISOLATION', 'true')
```
caseydavenport: I think calling this NAMESPACE_ISOLATION is a bit confusing, since namespace isolation will still be in effect if policy is declared, just not when no policy is declared.

A more apt name might be something like "DEFAULT_POLICY", with options for "ns_isolation", "allow".  This means we could potentially support multiple other default behaviors in the future.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -35,6 +35,9 @@
 KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
                                'http://kubernetes-master:8080/api/v1/')
 
+# Allow the user to enable/disable namespace isolation policy
+DEFAULT_POLICY = os.environ.get('DEFAULT_POLICY', 'ns_isolation')
```
caseydavenport: I think the default behavior should be no ns_isolation.

caseydavenport:
LGTM.  You'll want to update the calico-docker docs to reflect this as well.

https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md

PR #25, opened 2015-08-24
Namespace Kubernetes Annotation Key

alexhersh:
During the 8/14 Kubernetes Hangout, it was recommended that Kubernetes annotation keys be namespaced per organization. 
annotation key `policy` is now `projectcalico.org/policy`

TODO: Update calico-docker [doc](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesPolicy.md) to specify new key. These updates should be done in unison.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -24,6 +24,8 @@ DOCKER_VERSION = "1.16"
 ORCHESTRATOR_ID = "docker"
 HOSTNAME = socket.gethostname()
 
+POLICY_ANNOTATION_KEY = "projectcalico.org/policy"
+
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 if ETCD_AUTHORITY_ENV not in os.environ:
     os.environ[ETCD_AUTHORITY_ENV] = 'kubernetes-master:6666'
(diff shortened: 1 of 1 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -414,10 +414,10 @@ def _generate_rules(self, pod):
         annotations = self._get_metadata(pod, "annotations")
 
         # Find policy block of annotations
-        if annotations and "policy" in annotations.keys():
+        if annotations and "projectcalico.org/policy" in annotations.keys():
```
caseydavenport: Could you define this whole thing at the top as a POLICY_ANNOTATION_KEY=".."

calico_kubernetes/calico_kubernetes.py
```diff
@@ -414,10 +416,10 @@ def _generate_rules(self, pod):
         annotations = self._get_metadata(pod, "annotations")
 
         # Find policy block of annotations
-        if annotations and "policy" in annotations.keys():
+        if annotations and POLICY_ANNOTATION_KEY in annotations.keys():
```
caseydavenport: Another little nitpick.  You can check for the existence of a key in a dictionary without calling .keys().

This will do the same thing, without the extra work of calling keys()
```
if annotations and POLICY_ANNOTATION_KEY in annotations:
  ...
```

caseydavenport:
LGTM

Issue #29, opened 2015-08-28
Poor diagnostics when authentication with API server fails


caseydavenport:
Seen in #28 

When the plugin is configured to use secure comms with the API, but no BearerToken exists at `/var/lib/kubelet/kubernetes_auth`, the error is unclear.  The plugin should emit a clear log indicating the problem.

L-MA:
Fixed with PR #30

PR #30, opened 2015-08-28
Diagnostics when auth with API server fails

L-MA:
#29

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -379,16 +379,39 @@ class NetworkPlugin(object):
         :rtype: str
         """
         logger.info('Getting Kubernetes Authorization')
+
         try:
             with open('/var/lib/kubelet/kubernetes_auth') as f:
                 json_string = f.read()
         except IOError as e:
-            logger.info("Failed to open auth_file (%s), assuming insecure mode" % e)
-            return ""
+            if self._api_root_secure():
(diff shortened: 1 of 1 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -379,17 +379,40 @@ def _get_api_token(self):
...
                 json_string = f.read()
         except IOError as e:
-            logger.info("Failed to open auth_file (%s), assuming insecure mode" % e)
-            return ""
+            if self._api_root_secure():
+                logger.exception("Failed to open auth_file when API root is "
+                                 "set to secure (%s). Exiting", KUBE_API_ROOT)
+                sys.exit(1)
+            else:
+                logger.warning("No bearer token found")
```
caseydavenport: I'd combine these two logs into one - there's no need to have them on separate lines.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -379,17 +379,40 @@ def _get_api_token(self):
...
+                logger.warning("Using insecure mode")
+                return ""
 
         logger.info('Got kubernetes_auth: ' + json_string)
         auth_data = json.loads(json_string)
         return auth_data['BearerToken']
 
+    def _api_root_secure(self):
+        """
+        Checks whether the KUBE_API_ROOT is secure or insecure
```
caseydavenport: Make sure to use punctuation in all docstrings and comments!

PR #38, opened 2015-09-15
Policy Agent

alexhersh:


policy_agent/policy_agent.py
```diff
@@ -0,0 +1,354 @@
...
+            resource_pool = self.endpoints
+            obj = Endopints(target)
+
+        elif kind == "Namespace":
+            resource_pool = self.namespaces
+            obj = Namespace(target)
+
+        else:
+            _log.error("resource %s not Pod, Service, or Namespace" % kind)
+            sys.exit(1)
```
caseydavenport: I don't think we want to sys.exit() anywhere in this program - we just want to log an error and return.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,354 @@
...
+                             (target_key, target_key))
+                self.process_resource(action="ADDED", kind=kind, target=target)
+
+    def resync(self):
+        """
+        Tells all resource objects to resync their profile information
+        """
+        for resource_pool in [self.pods, self.services, self.endpoints, self.namespaces]:
+            for resource_key in resource_pool:
+                resource_pool[resource_key].resync()
```
caseydavenport: I don't think pushing the resync logic into the individual "resource" objects is the right move here.  It is going to make it difficult to combine information from multiple types of objects, and makes the resync code much harder to read and debug.

Let's talk about what the right abstraction is here.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    Resource objects pull pertinent info from json blobs and maintain universal functions  
+    """
+
+    def __init__(self, json):
+        """
+        On init, each Resource saves the raw json, pulls necessary info (unique), defines a unique key identifier, and sets self.needs_resync to True
+        """
+        self.json = json
+        self.from_json(json)
```
caseydavenport: This is not quite the pattern I was thinking of with "from_json" - it works fine (especially since we're only ever creating one of these from json) but in the more generic case, you'd want something like this:

```
class Thing()
  def __init__(self, param1, param2, param3):
    self.p1 = param1
    self.p2 = param2
    self.p3 = param3

  @classmethod
  def from_json(cls, json):
    # Returns a new instance
    return Thing(jsondata["p1"], jsondata["p2"], jsondata["p3"]
```

Then you can do:
```
thing = Thing.from_json(json_data)

OR

thing = Thing(p1, p2, p3)
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,354 @@
...
+        elif self.policy == "closed":
+            rules = Rules(id=ns_tag,
+                      inbound_rules=[Rule(action="allow", src_tag=ns_tag)],
+                      outbound_rules=[default_allow])
+            _log.info("Applying Closed Rules to NS Profile %s" % ns_tag)
+
+        else:
+            _log.error(
+                "Namespace %s policy is neither open nor closed" % self.name)
+            sys.exit(1)
```
caseydavenport: Remember this is a continuously running daemon.  If we sys.exit, it kills the process.

This is fine in the plugin, since it isn't continuously running and will be called again, but when we call sys.exit in here, we're terminating the entire process, which is bad.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    Called by watcher threads. Adds watch events to Queue
+    """
+    response = _get_api_stream("watch/%s" % path)
+    for line in response.iter_lines():
+        if line:
+            queue.put(line)
+
+
+def _get_api_token():
```
caseydavenport: Fine to copy this over for now, but we don't want to duplicate code like this in the long run - this should be common between the plugin and the agent.

caseydavenport:
Could you close this PR and open a new one against the `coarse-grained-policy` branch?

PR #39, opened 2015-09-17
Policy agent for coarse grained policy

L-MA:

PR #41, opened 2015-09-21
Policy agent support

L-MA:
This PR creates a Policy Agent that is responsible for implementing Calico policy on Kubernetes pods. Kubernetes calls the calico_kubernetes plugin, which creates an endpoint, creates a generic profile, and sets basic policy (if calico policy is utilized, deny all incoming traffic; otherwise, allow all traffic). 

The Policy Agent watches Kubernetes etcd for new resources and updates the pods accordingly. The intent is to implement the Kubernetes API experimental fields - `Namespace.Spec.experimentalNetworkPolicy` and `Service.Spec.Type=experimentalNamespaceIP`

Moving from #39 to pull against coarse-grained-policy branch


calico_kubernetes/calico_kubernetes.py
```diff
@@ -146,35 +143,29 @@ def _configure_profile(self, endpoint):
...
         """
-        pod = self._get_pod_config()
-
         logger.info('Configuring Pod Profile: %s', self.profile_name)
 
         if self._datastore_client.profile_exists(self.profile_name):
-            logger.error("Profile with name %s already exists, exiting.",
-                         self.profile_name)
-            sys.exit(1)
+            logger.warning("Profile with name %s already exists.",
```
caseydavenport: Aren't we going to hit this over and over again for each pod, since they're all going to be using the same profile?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -146,35 +143,29 @@ def _configure_profile(self, endpoint):
...
+            self._apply_rules()
 
         # Also set the profile for the workload.
         logger.info('Setting profile %s on endpoint %s',
                     self.profile_name, endpoint.endpoint_id)
-        self._datastore_client.set_profiles_on_endpoint(
-            [self.profile_name], endpoint_id=endpoint.endpoint_id
-        )
+
+        self._datastore_client.append_profiles_to_endpoint(profile_names=[self.profile_name],
```
caseydavenport: Why did we switch from set_profiles to append_profiles?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -182,16 +173,34 @@ def _configure_interface(self):
...
+        logger.info("Setting mac address %s to endpoint %s", endpoint.mac, endpoint.name)
+        self._datastore_client.set_endpoint(endpoint)
+
+        resource_path = "namespaces/%(namespace)s/pods/%(podname)s" % \
+                        {"namespace": self.namespace, "podname": self.pod_name}
+        ep_data = '{"metadata":{"annotations":{"%s":"%s"}}}' % \
+                  (EPID_ANNOTATION_KEY, endpoint.endpoint_id)
+        self._patch_api(path=resource_path, patch=ep_data)
+
+        interface_name = generate_cali_interface_name(IF_PREFIX, endpoint.endpoint_id)
```
caseydavenport: Isn't the endpoint name accessible from endpoint.name (or something like that?)

Do we really need to be generating it again here?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -203,14 +212,24 @@ def _configure_interface(self):
         # so we allocate an IP which is already allocated to the node. We set
         # the subnet to /32 so that the routing table is not affected;
         # no traffic for the node_ip's subnet will use the /32 route.
+        logger.info('Adding IP %s to interface %s', node_ip, interface_name)
         check_call(['ip', 'addr', 'add', node_ip + '/32',
                     'dev', interface_name])
-        logger.info('Finished configuring network interface')
-        return ep
+
+        # Give Kubernetes a link to the endpoint
```
caseydavenport: We're not really giving Kubernetes a link to it, we're annotating the endpoint ID on the pod object so that the policy agent can look-up the corresponding endpoint for this pod.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -356,9 +369,16 @@ def _get_container_info(self, container_name):
     def _get_container_pid(self, container_name):
         return self._get_container_info(container_name)["State"]["Pid"]
 
+    def _get_ipset_from_endpoint(self, endpoint):
```
caseydavenport: What was the reasoning for combining this into a helper method? It looks like there is only one place using this. 


I'd leave this as it was.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -356,9 +369,16 @@ def _get_container_info(self, container_name):
...
+        ip_set = set()
+        for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
+            ip_set.add(net.ip)
+        return ip_set
+
     def _read_docker_ip(self):
         """Get the IP for the pod's infra container."""
-        ip = self._get_container_info(self.docker_id)["NetworkSettings"]["IPAddress"]
+        ip = self._get_container_info(
```
caseydavenport: The formatting here is weird.  Either stick it on one line, or split like so:

```
info = self._get_container_info(self.docker_id)
ip = info["NetworkSettings"]["IPAddress"]
```

calico_kubernetes/calico_kubernetes.py
alexhersh: I think this line was there before endpoint.name existed. I will update.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -347,7 +359,8 @@ def _get_container_info(self, container_name):
             info = self._docker_client.inspect_container(container_name)
         except APIError as e:
             if e.response.status_code == 404:
-                logger.error("Container %s was not found. Exiting.", container_name)
+                logger.error(
```
caseydavenport: There are a number of logs that are formatted like this in this PR.

Could you update the formatting to look a bit nicer, like this for example:

```
logg.error("Container %s was not found. Exiting.", 
                  container_name)
```

or 

```
logger.error("This is a really long error message that probably" \
                    "needs to go on two lines")
```

calico_kubernetes/calico_kubernetes.py
alexhersh: Yes. Should we remove info message to cut down logging?

calico_kubernetes/calico_kubernetes.py
alexhersh: I was worried about a race condition where the policy agent would calculate policy before the plugin. Although, I suppose the default profile would still persist once it is caluclated...

calico_kubernetes/calico_kubernetes.py
```diff
@@ -454,7 +475,29 @@ def _get_api_path(self, path):
...
+    def _patch_api(self, path, patch):
+        """
+        Patch an api resource to a given path
+
+        :param path: The relative path to an API endpoint.
+        :param patch: The updated data
+        :return: A list of JSON API objects
+        :rtype list
+        """
+        logger.info(
```
caseydavenport: We don't need to print out KUBE_API_ROOT all over the place - we log it out once at the beginning and it never changes.

These logs can be reduced to a single debug level log which prints out the path and the patched data.

Info level logs can be added when this method is being called to provide higher-level context.

So, it might look like this:
```
logger.info("Patching pod with Calico endpoint ID")
self._patch_api(...)
```

calico_kubernetes/calico_kubernetes.py
caseydavenport: I'd add a debug level log in the profile_exists() branch that says something like "Profile %s already exists, no work to do" or something like that.

Add an info level log in the branch where the profile doesn't exist that says we're creating a new profile named X.

calico_kubernetes/calico_kubernetes.py
caseydavenport: The policy agent is not going to apply policy before the plugin does, because it relies on the endpoint ID annotation that the plugin patches on to the pod object.  Right?

I think the plugin should be using SET here to ensure the initial state is what we expect.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -497,217 +540,53 @@ def _api_root_secure(self):
                          'a http or https address. Exiting', KUBE_API_ROOT)
             sys.exit(1)
 
-    def _generate_rules(self, pod):
+    def _generate_profile_name(self):
```
caseydavenport: I think we can get rid of this.  The only reason this is still here is for the special case handling of the policy-agent pod, which we've determined is the wrong thing to do.

self.profile_name can be set at __init__ time.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -497,217 +540,53 @@ def _api_root_secure(self):
...
         """
-        Generate a rules for a given profile based on annotations.
-        1) Remove Calicoctl default rules
-        2) Create new profiles based on annotations, and establish new defaults
-
-        Exceptions:
-            If namespace = kube-system (internal kube services), allow all traffic
-            If no policy in annotations, allow from pod's Namespace
-            Outbound policy should allow all traffic
+        Generate a default rules
```
caseydavenport: The docstring should describe which rules will be applied.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -497,217 +540,53 @@ def _api_root_secure(self):
...
-        # Character to replace symbols
-        swap_char = '_'
+            default_rule = Rule(action="allow")
+            logger.info("Using allow all rules")
 
-        # If swap_char is in string, double it.
-        unescaped_string = re.sub(swap_char, "%s%s" % (swap_char, swap_char), unescaped_string)
+        rules = Rules(id=profile.name,
+                      inbound_rules=[default_rule],
+                      outbound_rules=[default_rule])
```
caseydavenport: As discussed, outbound_rule should always be "allow".

The CALICO_POLICY flag only controls the inbound rules.

policy_agent/calico-policy-rc.yaml
```diff
@@ -0,0 +1,32 @@
...
+        - name: KUBE_API_ROOT
+          value: "http://<KUBERNETES_MASTER>:8080/api/v1/"
+        - name: ETCD_AUTHORITY
+          value: "<KUBERNETES_MASTER>:4001"
+        resources:
+          limits:
+            cpu: 100m
+            memory: 50Mi
+        name: policyagent
+      dnsPolicy: Default  # Don't use cluster DNS.
```
caseydavenport: I think we can remove this line - it should be using cluster DNS.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+
+LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(filename)s: %(message)s'
+LOG_LEVEL = os.environ.get('LOG_LEVEL', 'INFO').upper()
+
+_log = logging.getLogger(__name__)
+_datastore_client = DatastoreClient()
+
+
+class PolicyAgent():
+
```
caseydavenport: No newlines before the docstring please :)

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+from subprocess import check_output
+from contextlib import closing
+from pycalico.datastore_datatypes import Rules, Rule
+from pycalico.datastore_errors import ProfileNotInEndpoint, ProfileAlreadyInEndpoint
+from pycalico.datastore import DatastoreClient, PROFILE_PATH
+
+KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
+                               'http://localhost:8080/api/v1/')
+
+ANNOTATION_NAMESPACE = "projectcalico.org/"
```
caseydavenport: This is common between the plugin and the agent, and should be defined in an external file which is imported into each so that we don't end up with mismatches.

A new `constants.py` would do the trick.

Then, the plugin and the agent could do:
```
from constants import ANNOTATION_NAMESPACE
```

calico_kubernetes/calico_kubernetes.py
alexhersh: Well we create the endpoint, then configure the profile, but there isnt much in between. I think you are right, I was just being overly cautious.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+EPID_ANNOTATION_KEY = "%sendpointID" % ANNOTATION_NAMESPACE
+
+POLICY_LOG_DIR = "/var/log/calico/policy"
+POLICY_LOG = "%s/calico.log" % POLICY_LOG_DIR
```
caseydavenport: I don't think we want to keep calling all of our logs `calico.log` but in a different location.

This should probably be "/var/log/calico/kubernetes/policy-agent.log"

calico_kubernetes/calico_kubernetes.py
L-MA: I broke this out when I was writing the CALICO_NETWORKING changes. Since its only getting called once I will get rid of this

calico_kubernetes/calico_kubernetes.py
caseydavenport: Let's make sure we kill the timing window completely by only annotating the Pod in the API after we've configured a profile on the endpoint.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        self.endpoints = dict()
+
+        self.changed_namespaces = dict()
+        self.changed_services = dict()
+        self.changed_pods = dict()
+        self.changed_endpoints = dict()
+
+    def run(self):
+        """
+        PolicyAgent.run() is called at program init to spawn watch threads and parse their responses
```
caseydavenport: Make sure you're using punctuation in your comments and docstrings!

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+
+        self.PodWatcher.start()
+        self.SvcWatcher.start()
+        self.NsWatcher.start()
+        self.EptsWatcher.start()
+
+        while True:
+            self.read_responses()
+
+    def init_lists(self):
```
caseydavenport: Could use a docstring.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        for svc in initServices["items"]:
+            self.process_resource(action="ADDED", kind="Service", target=svc)
+
+        for pod in initPods["items"]:
+            self.process_resource(action="ADDED", kind="Pod", target=pod)
+
+        for ep in initEndpoints["items"]:
+            self.process_resource(action="ADDED", kind="Endpoints", target=ep)
+
+        self.nsRV = initEndpoints["metadata"]["resourceVersion"]
```
caseydavenport: No need to abbreviate so much - make the code harder to figure out.

`nsRV` -> `nsResourceVersion`, etc.

caseydavenport: I'd declare this field in the __init__ as well and docstring it there so we know what it is.

```
def __init__(self):
  ...
  self.nsResourceVersion = None
  """
  The resourceVersion of the latest Namespace update we've received.
  """
```

calico_kubernetes/calico_kubernetes.py
alexhersh: I have a commit on my branch that fixesthis. After round 1 of comments I will have @L-MA merge it in.

calico_kubernetes/calico_kubernetes.py
alexhersh: Got it.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        self.poRV = initPods["metadata"]["resourceVersion"]
+        self.epRV = initNamespaces["metadata"]["resourceVersion"]
+
+        _log.info("Service Resource Version = %s" % self.svcRV)
+        _log.info("Namespace Resource Version = %s" % self.nsRV)
+        _log.info("Pod Resource Version = %s" % self.poRV)
+        _log.info("Endpoints Resource Version = %s" % self.epRV)
+
+    def init_threads(self):
+        self.NsWatcher = Thread(target=_keep_watch,
```
caseydavenport: Same with these guys - minimize abbreviations, and initialize them to `None` in your __init__.  In this method, you can assert that they are None to prevent accidentally calling this while threads are still running.

```
assert self.namespaceWatcher = None
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                                  args=(self.q, "endpoints", self.epRV))
+
+        self.NsWatcher.daemon = True
+        self.SvcWatcher.daemon = True
+        self.PodWatcher.daemon = True
+        self.EptsWatcher.daemon = True
+
+    def read_responses(self):
+        """
+        Read Responses pulls a response off the Queue and processes it
```
caseydavenport: Read Responses --> read_responses

caseydavenport: "pulls a response" -- what is a "response" in this case?

I think this method is actually "read_updates" - we aren't querying anything, so we're not receiving responses to a query.  We've just registered for updates.

caseydavenport: Remember punctuation :)

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        try:
+            response = self.q.get_nowait()
+            r_json = json.loads(response)
+            r_type = r_json["type"]
+            r_obj = r_json["object"]
+            r_kind = r_obj["kind"]
+
+            _log.info("%s: %s" % (r_type, r_kind))
+
+            if r_kind in ["Namespace", "Service", "Pod", "Endpoints"]:
```
caseydavenport: Generally nice to define these things as constants so you don't risk typos, etc screwing up your code.

```
KIND_NAMESPACE = "Namespace"
KIND_SERVICE = "Service"
KIND_POD = "Pod"
KIND_ENDPOINTS = "Endpoints"
VALID_KINDS = [ ... ]
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        If no responses remain in the queue, it will trigger a resync for all resource lists
+        """
+        try:
+            response = self.q.get_nowait()
+            r_json = json.loads(response)
+            r_type = r_json["type"]
+            r_obj = r_json["object"]
+            r_kind = r_obj["kind"]
+
+            _log.info("%s: %s" % (r_type, r_kind))
```
caseydavenport: I think this belongs as a debug log and should include more information.

Be wary of your logging here - this could potentially receive a lot of updates, thus cluttering the logs significantly.  Put things that might loop at debug level so they don't overwhelm.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+
+            if r_kind in ["Namespace", "Service", "Pod", "Endpoints"]:
+                if r_type == "DELETED":
+                    self.delete_resource(kind=r_kind, target=r_obj)
+                elif r_type in ["ADDED", "MODIFIED"]:
+                    self.process_resource(
+                        action=r_type, kind=r_kind, target=r_obj)
+                else:
+                    _log.error("Event from watch not recognized: %s" % r_type)
+
```
caseydavenport: Extra newline here.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                            _log.warning("Pod %s is not yet processed" % (pod))
+
+                # declare Endpoints obj processed
+                self.epRV = ep.RV
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+
+
+class Resource():
+
```
caseydavenport: Extra newline before docstring.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        self.json = json
+        self.RV = json["metadata"]["resourceVersion"]
+        self.from_json(json)
+        self.key = self.get_key()
+
+    def from_json(self, json):
+        self.name = "noop"
+        return
+
+    def get_key(self):
```
caseydavenport: This is meant to be overridden - any subclass of "Resource" must override this method.

Instead of providing a default, this should raise `NotImplementedError` to indicate that the subclass hasn't overridden this method.

caseydavenport: Same with `from_json`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+    """
+    Resource objects pull pertinent info from json blobs and maintain universal functions
+    """
+
+    def __init__(self, json):
+        """
+        On init, each Resource saves the raw json, pulls necessary info (unique),
+        and defines a unique key identifier
+        """
+        self.json = json
```
caseydavenport: `self.json` should probably be private --> i.e renamed to `self._json`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        self.RV = json["metadata"]["resourceVersion"]
```
caseydavenport: unnecessary abbreviation and non-standard naming (don't use all-caps for field names).

This should be `self.resourceVersion`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        self.from_json(json)
+        self.key = self.get_key()
```
caseydavenport: No need to assign `self.key` in __init__

If you use the `@property` decorator, you can simply rename and use your `get_key` for the same purpose.

So,
```
@property
def key(self):
  """
  Returns a key using fields a, b and c.
  """
  return (a, b, c)
```

can then be used like this:

```
k = some_resource.key
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+    def get_key(self):
+        return self.name
+
+    def __str__(self):
+        return "%s: %s\n%s" % (self.kind, self.key, self.json)
+
+
+class Namespace(Resource):
+
+    def from_json(self, json):
```
caseydavenport: I wouldn't call this method `from_json` the way you have it right now.

`from_json` implies you're using a `classmethod` to return an instance of this class.  The way you're using is more of a `init_from_json` or something like that.

You should either rename it, or refactor this to use the proper factory `classmethod`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        try:
+            self.policy = json["spec"]["experimentalNetworkPolicy"]
+        except KeyError:
+            _log.warning("Namespace does not have policy, assumed Open")
+            self.policy = "Open"
+
+    def get_key(self):
+        return self.name
+
+    def create_ns_profile(self):
```
caseydavenport: Needs a docstring.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+            return False
+
+        # Write rules to profile
+        _datastore_client.etcd_client.write(
+            profile_path + "rules", rules.to_json())
+        return True
+
+    def delete_ns_profile(self):
+        # Derive NS tag
+        ns_tag = "namespace_%s" % self.name
```
caseydavenport: `ns_tag` is actaully the profile_name, correct?

As such, it should probably be called profile name instead.  It would probably make sense to create the name in `__init__` as self.profile_name so it can be referenced later.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+
+
+class Service(Resource):
+
+    def from_json(self, json):
+        self.kind = "Service"
+        self.name = json["metadata"]["name"]
+        self.namespace = json["metadata"]["namespace"]
+        try:
+            self.type = json["spec"]["type"]
```
caseydavenport: Do we ever expect the type to not be defined?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        except KeyError:
+            _log.warning("Service Type not specified assuming NamespaceIP")
+            self.type = "NamespaceIP"
```
caseydavenport: `NamespaceIP` probably warrants a constant for ease of use.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                pass
+            return True
+        else:
+            _log.error("Pod Resource %s found before Namespace Resource %s" % (
+                self.name, self.namespace))
+            return False
+
+    def remove_default_profile(self):
+        """
+        remove the default reject all rule programmed by the plugin
```
caseydavenport: Punctuation and capitalization.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        """
+        default_profile = "REJECT_ALL"
```
caseydavenport: This should be commonized into a constant somewhere.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                rules = Rules(id=profile_name,
+                              inbound_rules=port_rules,
+                              outbound_rules=[default_allow])
+
+                # Write rules to profile
+                _datastore_client.etcd_client.write(
+                    profile_path + "rules", rules.to_json())
+
+            else:
+                _log.warning("Profile %s already in Store" % profile_name)
```
caseydavenport: What does the profile store sell?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                dst_port = verify_args(port_spec, "port")
+                protocol = verify_args(port_spec, "protocol").lower()
+                if protocol and dst_port:
+                    port_rules.append(Rule(action="allow",
+                                           dst_ports=[dst_port],
+                                           protocol=protocol))
+
+            # Create Profile for the subset
+            # TODO: better profile names
+            profile_name = "svc_%s_%s_port_%s_protocol_%s" % (
```
caseydavenport: Yep - we should fix the naming up here.

It looks like there are potentially multiple port/protocol combinations that could be exposed in a port spec.

So, an appropriate name might be: `%ns_svc_%svc_%protocol1%port1_...%protocolN%portN"`

Which might look something like this in practice: `production_svc_frontend_tcp80_udp80`

calico_kubernetes/calico_kubernetes.py
alexhersh: Done

calico_kubernetes/calico_kubernetes.py
alexhersh: Done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

policy_agent/calico-policy-rc.yaml
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
L-MA: done

policy_agent/policy_agent.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
L-MA: done

calico_kubernetes/calico_kubernetes.py
L-MA: done

calico_kubernetes/calico_kubernetes.py
L-MA: done

policy_agent/policy_agent.py
alexhersh: semi-done

calico_kubernetes/calico_kubernetes.py
L-MA: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+
+                # Write rules to profile
+                _datastore_client.etcd_client.write(
+                    profile_path + "rules", rules.to_json())
+
+            else:
+                _log.warning("Profile %s already in Store" % profile_name)
+
+            # Generate list of pod names
+            pods = list()
```
caseydavenport: As a general note - Lists and dictionaries are usually declared like so:
```
somelist = []
dictionary = {}
```

I don't think there is anything wrong with the way you are doing it, but it sure looks weird to me. ;)

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: using `init_from_json` for now

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                    queue.put(line)
+        except:
+            # If we hit an exception attempting to watch this path, log it, and retry the watch
+            # after a short sleep in order to prevent tight-looping.  We catch all BaseExceptions
+            # so that the thread never dies.
+            _log.exception("Exception watching path %s", resource)
+            time.sleep(10)
+
+
+def _get_api_token():
```
caseydavenport: More common function between the plugin and the policy agent.   Don't want to duplicate code, so this should be moved to a central location accessible by both.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        'Streaming API Resource: %s from KUBE_API_ROOT: %s', path, KUBE_API_ROOT)
+    bearer_token = _get_api_token()
+    session = requests.Session()
+    session.headers.update({'Authorization': 'Bearer ' + bearer_token})
+    session = requests.Session()
+    return session.get("%s%s" % (KUBE_API_ROOT, path),
+                       verify=False, stream=True)
+
+
+def _get_api_list(resource):
```
caseydavenport: I think this code also exists in the plugin - should move to a common location.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+            response = self.q.get_nowait()
+            r_json = json.loads(response)
+            r_type = r_json["type"]
+            r_obj = r_json["object"]
+            r_kind = r_obj["kind"]
+
+            _log.info("%s: %s" % (r_type, r_kind))
+
+            if r_kind in ["Namespace", "Service", "Pod", "Endpoints"]:
+                if r_type == "DELETED":
```
caseydavenport: DELETED, ADDED, and MODIFIED are also used enough to warrant constants.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+        # As endpoint lists change, create/add new profiles to pods as
+        # necessary
+        # TODO: if some object not ready for resync, finish loop w/out moving to
+        # processed list
+        for ep_key, ep in self.changed_endpoints.items():
+
+            # Get policy of the Service and namespace. Uses same namespace/name
+            # dict key.
+            try:
+                svc_policy = self.services[ep_key].type
```
caseydavenport: svc_policy -> svc_type

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+            except KeyError:
+                _log.warning("Service %s not yet in store" % ep_key)
+                continue
+
+            try:
+                ns_policy = self.namespaces[ep.namespace].policy
+            except KeyError, AttributeError:
```
caseydavenport: What throws the AttributeError?  That seems like a different error than the Namespace not being in the datastore yet.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+                _log.warning("Namespace %s not yet in store" % ep.namespace)
+                continue
+
+            _log.info("Using svc policy %s and ns policy %s" %
+                      (svc_policy, ns_policy))
+
+            # if namespace is open, or svc is closed, do nothing
```
caseydavenport: You should elaborate on why this is true here.

policy_agent/policy_agent.py
L-MA: Nope, the default is always set to ClusterIP

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+            if ns_policy == "Open" or svc_policy == "NamespaceIP":
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+
+            # else define service policy
```
caseydavenport: Don't stick newlines or comments between your if / else blocks.  In general commenting should look something like this:

```
# Make action based on whether or not certain condition is true because A, B, C. (This one is optional)
if condition:
  # The condition is true, we must do X because blah.
  do_x()
else:
  # The condition is not true, we must do Y because blah.
  do_y()
```

policy_agent/policy_agent.py
alexhersh: way ahead of you. done.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,643 @@
...
+            else:
+                # find pods and append profiles
+                associations = ep.generate_svc_profiles_pods()
+                for profile, pods in associations.items():
+                    for pod in pods:
+                        existing_pod = self.match_pod(
+                            namespace=ep.namespace, name=pod)
+                        #_log.info("Adding profile %s to pod %s" % (profile, existing_pod))
```
caseydavenport: Whats up with this commented out log?  Too spammy?  If so, let's remove it or demote it to debug level.

alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: That was a problem I fixed at both ends. Tidying up. (done)

policy_agent/policy_agent.py
L-MA: done

policy_agent/policy_agent.py
L-MA: done

policy_agent/policy_agent.py
L-MA: done

policy_agent/policy_agent.py
L-MA: done

policy_agent/policy_agent.py
caseydavenport: How is it semi-done?

policy_agent/policy_agent.py
caseydavenport: Then we should get rid of the whole try/except thing around this.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -37,16 +36,14 @@
...
-                               'http://kubernetes-master:8080/api/v1/')
-
-# Allow the user to enable/disable namespace isolation policy
-DEFAULT_POLICY = os.environ.get('DEFAULT_POLICY', 'allow')
-
 # Flag to indicate whether or not to use Calico IPAM.
 # If False, use the default docker container ip address to create container.
 # If True, use libcalico's auto_assign IPAM to create container.
-CALICO_IPAM = os.environ.get('CALICO_IPAM', 'false')
+CALICO_IPAM = os.environ.get('CALICO_IPAM', 'true')
```
caseydavenport: CALICO_IPAM shouldn't default to true.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -82,26 +80,25 @@ def create(self, namespace, pod_name, docker_id):
                          e.returncode, e.output, e)
             sys.exit(1)
 
+        # Give the policy agent a reference to the endpoint in the API
```
caseydavenport: Add a comment explaining that this must be done after configuring the default profile to avoid the race condition we talked about.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -82,26 +80,25 @@ def create(self, namespace, pod_name, docker_id):
...
     def delete(self, namespace, pod_name, docker_id):
         """Cleanup after a pod."""
         self.pod_name = pod_name
         self.docker_id = docker_id
         self.namespace = namespace
-        self.profile_name = "%s_%s_%s" % (self.namespace, self.pod_name, str(self.docker_id)[:12])
 
-        logger.info('Deleting container %s with profile %s',
-                    self.docker_id, self.profile_name)
+        logger.info('Deleting container %s',
```
caseydavenport: This log can be on a single line.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -203,14 +204,18 @@ def _configure_interface(self):
...
+        node_ip = self._get_node_ip()
+        logger.info('Adding IP %s to interface %s', node_ip, endpoint.name)
         check_call(['ip', 'addr', 'add', node_ip + '/32',
-                    'dev', interface_name])
-        logger.info('Finished configuring network interface')
-        return ep
+                    'dev', endpoint.name])
+
+        logger.info('Finished configuring Calico network interface')
+
```
caseydavenport: Extra newline here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -182,16 +174,25 @@ def _configure_interface(self):
...
         self._delete_docker_interface()
-        logger.info('Configuring Calico network interface')
-        ep = self._container_add(container_pid, interface)
-        interface_name = generate_cali_interface_name(IF_PREFIX, ep.endpoint_id)
-        node_ip = self._get_node_ip()
-        logger.info('Adding IP %s to interface %s', node_ip, interface_name)
+
+        # Create the veth, move into the container namespace, add the IP and
+        # set up the default routes.
+        logger.info("Creating the veth with namespace pid %s on interface name %s", container_pid, interface)
```
caseydavenport: I think this block of code should be returned to _container_add()

calico_kubernetes/calico_kubernetes.py
```diff
@@ -266,18 +264,17 @@ def _assign_container_ip(self):
...
         if CALICO_IPAM == 'true':
-            # Assign ip address through IPAM Client
+            # Assign IP address through IPAM Client
             logger.info("Using Calico IPAM")
             try:
                 ip_list, ipv6_addrs = self._datastore_client.auto_assign_ips(
                     1, 0, self.docker_id, None)
                 ip = ip_list[0]
-                logger.debug("ip_list is %s; ipv6_addrs is %s", ip_list, ipv6_addrs)
+                logger.debug( "ip_list is %s; ipv6_addrs is %s", ip_list, ipv6_addrs)
```
caseydavenport: Rogue space at the beginning of this log.

common/constants.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT', 'http://localhost:8080/api/v1/')
```
caseydavenport: These should all have comments explaining what they mean / configure.

Things that are part of "groups" (e.g. logging constants, DEFAULT_PROFILE_*, annotation keys) should be grouped accordingly using newlines and comments to explain the block.

policy_agent/policy_agent.py
caseydavenport: Still have a newline here...

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        self.pod_watcher.daemon = True
+        self.endpoints_watcher.daemon = True
+
+    def read_updates(self):
+        """
+        Pulls an update off the Queue and processes it.
+        If no responses remain in the queue, it will trigger a resync for all resource lists.
+        """
+        try:
+            response = self.watcher_queue.get_nowait()
```
caseydavenport: If there are no responses in the queue for a while, I think this will tight-loop and keep attempting to resync.

I think we can just use self.watcher_queue.get() so that we wait for the queue to be populated.  This will prevent tight-looping when the queue is empty.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            r_json = json.loads(response)
+            r_type = r_json["type"]
+            r_obj = r_json["object"]
+            r_kind = r_obj["kind"]
+
+            _log.debug("Reading update %s: %s" % (r_type, r_kind))
+
+            if r_kind in VALID_KINDS:
+                if r_type == CMD_DELETED:
```
caseydavenport: Mismatch in terminology here - Is it a type, or a command?

I think this should be `cmd` instead of `r_type` - it's important to keep your naming consistent for more readable code.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def read_updates(self):
+        """
+        Pulls an update off the Queue and processes it.
+        If no responses remain in the queue, it will trigger a resync for all resource lists.
+        """
+        try:
+            response = self.watcher_queue.get_nowait()
+            r_json = json.loads(response)
+            r_type = r_json["type"]
```
caseydavenport: I'd remove the `r_` prefix from all of these - it doesn't add anything, and makes it slightly harder to read.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                if r_type == CMD_DELETED:
+                    self.delete_resource(kind=r_kind, target=r_obj)
+                elif r_type in [CMD_ADDED, CMD_MODIFIED]:
+                    self.process_resource(action=r_type,
+                                          kind=r_kind,
+                                          target=r_obj)
+                else:
+                    _log.error("Event from watch not recognized: %s" % r_type)
+            else:
+                _log.error("Resource %s not Pod, Service, Endpoints or Namespace" % r_kind)
```
caseydavenport: You can use VALID_KINDs here instead of hard-coding the valid kinds in the log message.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        elif kind == KIND_SERVICE:
+            resource_pool = self.changed_services
+            obj = Service(target)
+        elif kind == KIND_POD:
+            resource_pool = self.changed_pods
+            obj = Pod(target)
+        elif kind == KIND_ENDPOINTS:
+            resource_pool = self.changed_endpoints
+            obj = Endpoints(target)
+        else:
```
caseydavenport: Rather than have an else branch here, let's just assert that kind is in VALID_KINDS at the beginning of this method.  If we pass an invalid kind here, that is a bug in our code, so we should fail loudly.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            _log.error("Resource %s not Pod, Service, Endpoints or Namespace" % kind)
+            return
+
+        target_key = obj.key
```
caseydavenport: Unnecessary masking of obj.key.  obj.key is actually fewer characters than target_key, so why store it in a local variable?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        """
+        Takes a target object and an action and updates internal resource pools
+        :param action: String of [CMD_ADDED, CMD_MODIFIED] (returned by api watch)
+        :param kind: String in VALID_KINDS
+        :param target: json dict of resource info
+        """
+        # Determine Resource Pool
+        if kind == KIND_NAMESPACE:
+            resource_pool = self.changed_namespaces
+            obj = Namespace(target)
```
caseydavenport: Another instance of naming inconsistency - we're calling them "resources" everywhere else, so why call it "obj" here?

`obj` should be renamed to `resource` (Although I'm not entirely happy with the term "resource" itself - its a bit too generic)

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                                          kind=r_kind,
+                                          target=r_obj)
+                else:
+                    _log.error("Event from watch not recognized: %s" % r_type)
+            else:
+                _log.error("Resource %s not Pod, Service, Endpoints or Namespace" % r_kind)
+        except Queue.Empty:
+            self.resync()
+
+    def process_resource(self, action, kind, target):
```
caseydavenport: `target` should be named `resource`

caseydavenport: `action` should be named `command`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        :param action: String of [CMD_ADDED, CMD_MODIFIED] (returned by api watch)
+        :param kind: String in VALID_KINDS
+        :param target: json dict of resource info
+        """
+        # Determine Resource Pool
+        if kind == KIND_NAMESPACE:
+            resource_pool = self.changed_namespaces
+            obj = Namespace(target)
+        elif kind == KIND_SERVICE:
+            resource_pool = self.changed_services
```
caseydavenport: I'd rename `resource_pool` to something like `changed_resources` - its more descriptive.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                # When the Namespace is open, all valid service types are open as well.
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+            else:
+                # else define service policy
+                associations = ep.generate_svc_profiles_pods()
+                for profile, pods in associations.items():
+                    # find pods and append profiles
+                    for pod in pods:
+                        existing_pod = self.match_pod(
```
caseydavenport: I don't think you need the whole self.match_pod(..) helper.

Just do this:
```
existing_pod = self.pods.get((ep.namespace, pod_name))
```

If the given key doesn't exist, `get()` will return `None`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                # in which case NamespaceIP service policy is redundant
+                # When the Namespace is open, all valid service types are open as well.
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+            else:
+                # else define service policy
+                associations = ep.generate_svc_profiles_pods()
+                for profile, pods in associations.items():
+                    # find pods and append profiles
+                    for pod in pods:
```
caseydavenport: Confusing naming here - you're actually iterating through pod names, not pods.

We use the term "pod" to represent a Pod object.

As such, this should be:
```
for pod_name in pods_names:
  ...
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                addresses: [
+                    ...pods...
+                ],
+                ports: [
+                    ...associated ports...
+                ]
+            },
+        ]
+        :return: A generated dict of profile-pod associations
+        :rtype: a dict of profiles mapping to a list of assoc pods
```
caseydavenport: "assoc" - no need to abbreviate this.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        self.namespace = json["metadata"]["namespace"]
+        self.subsets = json["subsets"]
+
+    @property
+    def key(self):
+        return "%s/%s" % (self.namespace, self.name)
+
+    def generate_svc_profiles_pods(self):
+        """
+        Endpoints objects contain a list of subsets
```
caseydavenport: weird formatting here - this should be one line.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        :rtype: a dict of profiles mapping to a list of assoc pods
+        """
+        def verify_args(ports_obj, try_arg):
+            try:
+                arg = ports_obj[try_arg]
+            except KeyError:
+                arg = None
+            return arg
+
+        self.association_list = {}
```
caseydavenport: why is this a field on the object, rather than a local variable?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                ],
+                ports: [
+                    ...associated ports...
+                ]
+            },
+        ]
+        :return: A generated dict of profile-pod associations
+        :rtype: a dict of profiles mapping to a list of assoc pods
+        """
+        def verify_args(ports_obj, try_arg):
```
caseydavenport: This is re-inventing Dict.get(...) - no need for this guy.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        self.association_list = {}
+        for subset in self.subsets:
+            # Base profile_name string
+            profile_name = "%s_svc_%s" % (self.namespace, self.name)
+
+            # Generate a list of new svc profiles per port spec
+            port_rules = []
+            for port_spec in subset["ports"]:
+                dst_port = verify_args(port_spec, "port")
+                protocol = verify_args(port_spec, "protocol").lower()
```
caseydavenport: What happens when `verify_args()` returns `None`, and you try to call `.lower()` on it?

caseydavenport: Will there ever be a port_spec that doesn't contain a port an protocol? Do we even need to handle the `None` case?

I would think not.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        def verify_args(ports_obj, try_arg):
+            try:
+                arg = ports_obj[try_arg]
+            except KeyError:
+                arg = None
+            return arg
+
+        self.association_list = {}
+        for subset in self.subsets:
+            # Base profile_name string
```
caseydavenport: Missing punctuation on all of these comments.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            # Generate a list of new svc profiles per port spec
+            port_rules = []
+            for port_spec in subset["ports"]:
+                dst_port = verify_args(port_spec, "port")
+                protocol = verify_args(port_spec, "protocol").lower()
+                if protocol and dst_port:
+                    port_rules.append(Rule(action="allow",
+                                           dst_ports=[dst_port],
+                                           protocol=protocol))
+                    profile_name = "%s_%s%s" % (
```
caseydavenport:
```
profile_name += "_%s%s" % (protocol, dst_port)
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                        profile_name, protocol, dst_port)
+
+            # Create Profile for the subset
+            if not _datastore_client.profile_exists(profile_name):
+                _log.info("Creating Profile %s" % profile_name)
+                profile_path = PROFILE_PATH % {"profile_id": profile_name}
+                _datastore_client.etcd_client.write(
```
caseydavenport: Is there no libcalico function to add tags to a profile without directly accessing etcd?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                    profile_path + "tags", '["%s"]' % profile_name)
+
+                # Determine rule set
+                default_allow = Rule(action="allow")
+                rules = Rules(id=profile_name,
+                              inbound_rules=port_rules,
+                              outbound_rules=[default_allow])
+
+                # Write rules to profile
+                _datastore_client.etcd_client.write(
```
caseydavenport: Same thing with rules - is there no way to do this using libcalico?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                    profile_path + "rules", rules.to_json())
+            else:
+                _log.warning("Profile %s already exists" % profile_name)
+
+            # Generate list of pod names
+            pods = []
```
caseydavenport: We need to be resilient to the targetRef not existing - this won't exist for all services.

If it doesn't exist its fine - we just don't apply policy on those endpoints - but we definitely can't crash due to a `KeyError`

policy_agent/policy_agent.py
caseydavenport: This can be removed and the one in `util.py` can be used.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+
+import requests
+
+from common.constants import *
+from pycalico.datastore_datatypes import Rules, Rule
+from pycalico.datastore_errors import ProfileNotInEndpoint, ProfileAlreadyInEndpoint
+from pycalico.datastore import DatastoreClient, PROFILE_PATH
+
+POLICY_LOG_DIR = "/var/log/calico/policy"
+POLICY_LOG = "%s/calico.log" % POLICY_LOG_DIR
```
caseydavenport: This log should be renamed to policy-agent.log

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+from threading import Thread
+
+import requests
+
+from common.constants import *
+from pycalico.datastore_datatypes import Rules, Rule
+from pycalico.datastore_errors import ProfileNotInEndpoint, ProfileAlreadyInEndpoint
+from pycalico.datastore import DatastoreClient, PROFILE_PATH
+
+POLICY_LOG_DIR = "/var/log/calico/policy"
```
caseydavenport: I think this log should be under kubernetes, rather than "policy".

This agent is Kubernetes specific.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+    session = requests.Session()
+    return session.get("%s%s" % (KUBE_API_ROOT, path),
+                       verify=False, stream=True)
+
+
+def _get_api_list(resource):
+    """
+    Get a resource from the API specified API path.
+    e.g.
+    _get_api_path(default, service, nginx)
```
caseydavenport: docstring is wrong.

policy_agent/policy_agent.py
alexhersh: the write tags line is essentially what the datastore client uses to create a profile. I didn't want to use create_profile because it creates a default set of rules... however I did find a more efficient pycalico function to use so Ill throw it in

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+
+        for service_key, svc in self.changed_services.items():
+            # Process Services
+            self.svcResourceVersion = svc.resourceVersion
+            self.services[service_key] = svc
+            del self.changed_services[service_key]
+
+        for pod_key, po in self.changed_pods.items():
+            # Apply Namespace policy to new/updated pods
+            if po.apply_ns_policy():
```
caseydavenport: I'm not a super big fan of these methods which perform an action (such as apply ns policy) and then return True / False based on their success - it isn't very Pythonic.

Methods like "apply_ns_profile" which perform a very specific task are also a bit of a turn off, since they're not very flexible or re-usable.

I'd expect this API to look more like this:
```
for pod_key, pod in self.changed_pods.iteritems():
  namespace = self.namespaces.get(pod.namespace)
  if not namespace:
    # Namespace for this pod does not yet exist - skip.  It will retain the default REJECT_ALL profile.
    _log.warning(...)
    continue
  
  # The namespace exists.  Set the profiles - this also removes any existing profiles on the pod.
  ns_profile = namespace.profile
  pod.set_profiles([ns_profile])

  # Update our internal state
  self.podResourceVersion = pod.resourceVersion
  self.pods[pod.key] = pod
  del self.changed_pods[pod.key]
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                if po.remove_default_profile():
+                    self.poResourceVersion = po.resourceVersion
```
caseydavenport: Could you add the "d" on the end of pod pleeease...

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            except KeyError:
+                _log.warning("Namespace %s not yet in store" % ep.namespace)
+                continue
+
+            _log.info("Using svc policy %s and ns policy %s" % \
+                      (svc_type, ns_policy))
+
+            if ns_policy == "Open" or svc_type == SVC_TYPE_NAMESPACE_IP:
+                # If namespace is open, or svc is closed, do nothing.
+                # Namespace_IP services can only exists in Closed namespaces,
```
caseydavenport: `Namespace_IP` -> `NamespaceIP`

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                # in which case NamespaceIP service policy is redundant
+                # When the Namespace is open, all valid service types are open as well.
```
caseydavenport: This branch needs a log.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            try:
+                ns_policy = self.namespaces[ep.namespace].policy
+            except KeyError:
+                _log.warning("Namespace %s not yet in store" % ep.namespace)
+                continue
+
+            _log.info("Using svc policy %s and ns policy %s" % \
+                      (svc_type, ns_policy))
+
+            if ns_policy == "Open" or svc_type == SVC_TYPE_NAMESPACE_IP:
```
caseydavenport: No constant for "Open" ?

policy_agent/policy_agent.py
alexhersh: I need it for reference when deleting profiles. Although I suppose it makes returning redundant

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                # If namespace is open, or svc is closed, do nothing.
+                # Namespace_IP services can only exists in Closed namespaces,
+                # in which case NamespaceIP service policy is redundant
+                # When the Namespace is open, all valid service types are open as well.
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+            else:
+                # else define service policy
```
caseydavenport: Capitalization and punctuation!

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+                associations = ep.generate_svc_profiles_pods()
+                for profile, pods in associations.items():
+                    # find pods and append profiles
+                    for pod in pods:
+                        existing_pod = self.match_pod(
+                            namespace=ep.namespace, name=pod)
+                        _log.debug("Adding profile %s to pod %s" % (profile, existing_pod))
+                        if existing_pod:
```
caseydavenport: This could be encapsulated inside the pod object for code re-use.  See my comment above regarding the API, but it might look like this:
```
pod.append_profile(...)
pod.set_profiles(...)
pod.remove_profile(...)
```

policy_agent/policy_agent.py
alexhersh: I use match_pod more for the conversion of namespace and name strings into a pod object. I use that pattern a few times and I wanted to unify it so there isn't a mismatch. I could simplify it into `get_pod_key(namespace, name)` ?

caseydavenport: I only see `match_pod` called once, and `match-pod` in itself is really just a re-write of `Dict.get()`.  I'm not sure I understand what benefit it is providing.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+        try:
+            return self.pods["%s/%s" % (namespace, name)]
+        except KeyError:
+            _log.error("Pod %s in NS %s not processed" % (name, namespace))
+            return None
+
+    def resync(self):
+        """
+        Updates Calico store with new resource information
+        """
```
caseydavenport: Resync needs a nice info level log to indicate we've started a re-sync.

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            # Apply Namespace policy to new/updated pods
+            if po.apply_ns_policy():
+                if po.remove_default_profile():
+                    self.poResourceVersion = po.resourceVersion
+                    self.pods[pod_key] = po
+                    del self.changed_pods[pod_key]
+
+        # As endpoint lists change, create/add new profiles to pods as
+        # necessary
+        # TODO: if some object not ready for resync, finish loop w/out moving to
```
caseydavenport: I think this TODO has been done?

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+
+                # declare Endpoints obj processed
+                self.epResourceVersion = ep.resourceVersion
+                self.endpoints[ep_key] = ep
+                del self.changed_endpoints[ep_key]
+
+
+class Resource():
+    """
+    Resource objects pull pertinent info from json blobs and maintain universal functions
```
caseydavenport: What does "maintain universal functions" mean here?  Does it mean they have methods?  That is sort of implied ;)

A more useful docstring for the Resource class might be:

```
"""
The Resource class is an abstract super class which represents Kubernetes API objects.  The class defines a number of abstract methods which must be implemented in subclasses.
"""
```

policy_agent/policy_agent.py
```diff
@@ -0,0 +1,655 @@
...
+            self.policy = json["spec"]["experimentalNetworkPolicy"]
+        except KeyError:
+            _log.warning("Namespace does not have policy, assumed Open")
+            self.policy = "Open"
+
+    @property
+    def key(self):
+        return self.name
+
+    def create_ns_profile(self):
```
caseydavenport: `create_ns_profile` -> `create_profile`

The "ns" bit is implied, since we're operating on a namespace object.

policy_agent/policy_agent.py
alexhersh: If an endpoints obj is processed before a pod within it, it will not apply policy but consider the endpoints processed.

caseydavenport: Is this possible? Will we receive an endpoints which references a pod that hasn't been created?

I would think we can guarantee that we'll receive a pod update before we receive an endpoints update - so long as we process changed pods before endpoints, we'd be ok.  Shout if this is wrong, because it will probably require some deeper thinking.

policy_agent/policy_agent.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: If the endpoints watcher connects to the API before the pod watcher, the resources will be placed in the queue out of the assumed order. Logically changes to pods and services will trigger a subsequent Endpoints change, but our Threads are pulling updates asynchronously, so who knows what order things show up in queues?

Also resync may not always process everything in one go. If something in pod processing fails, it will skip and retry the next cycle.

common/constants.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: I am using resource because that is what Kubernetes uses to refer to [pods, services, replication controllers, etc...] 
(replaced obj with resource)

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: removed return, kept obj field

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done. As Alton Brown has said on many an occasion: The only single-use tool you need is a fire extinguisher

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: constants added

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: done

policy_agent/policy_agent.py
alexhersh: Ok it is gone

policy_agent/policy_agent.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: done

calico_kubernetes/calico_kubernetes.py
alexhersh: ignoring for this rev

caseydavenport:
Closing in favor of #43

Issue #8, opened 2015-08-12
Labels, Profiles, Tags should be properly escaped


caseydavenport:
Labels, tags, and profile names are passed to Felix which use them as rule-names (for human readability). However, Kubernetes accepts the `/` character (as well as a few others), which Felix will error on. Therefore, we need to properly escape these characters.

We've temporarily worked around this issue by replacing incompatible characters with underscores.  However, there is the potential for clashes between escaped strings, so more robust escaping is warranted.

Marking as low priority, since these clashes (though possible) are pretty unlikely in real-world scenarios.

PR #53, opened 2015-09-29
Move policy from kubernetes plugin to CNI plugin

L-MA:
[Review on Reviewable](https://reviewable.io/reviews/projectcalico/calico-kubernetes/53)

### DIFF
--- a/calico_kubernetes/calico_kubernetes_cni.py
+++ b/calico_kubernetes/calico_kubernetes_cni.py
@@ -110,6 +110,9 @@ def delete(args):
     """
     container_id = args['container_id']
     net_name = args['name']
+    namespace = args['namespace']
+    pod_name = args['pod_name']
+    profile_name = '%s_%s_%s' % (namespace, pod_name, container_id[:12])
 
     _log.info('Deleting pod %s' % container_id)
 
@@ -119,11 +122,11 @@ def delete(args):
                       container_id=container_id)
 
     # Delete profile if only member
-    if _datastore_client.profile_exists(net_name) and \
-       len(_datastore_client.get_profile_members(net_name)) < 1:
+    if _datastore_client.profile_exists(profile_name) and \
+       len(_datastore_client.get_profile_members(profile_name)) < 1:
         try:
-            _log.info("Profile %s has no members, removing from datastore" % net_name)
-            _datastore_client.remove_profile(net_name)
+            _log.info("Profile %s has no members, removing from datastore" % profile_name)
+            _datastore_client.remove_profile(profile_name)
         except:
             _log.error("Cannot remove profile %s: Profile cannot be found." % container_id)
             sys.exit(1)
@@ -251,7 +254,7 @@ def _set_profile_on_endpoint(container_id, namespace, endpoint, pod_name, profil
 
     # Also set the profile for the workload.
     _datastore_client.set_profiles_on_endpoint(profile_names=[profile_name],
-                                              endpoint_id=endpoint.endpoint_id)
+                                               endpoint_id=endpoint.endpoint_id)
 
 
 def _assign_to_pool(subnet):
@@ -381,8 +384,6 @@ def _apply_tags(container_id, namespace, pod_name, profile_name):
             tag = _label_to_tag(k, v)
             _log.info('Adding tag ' + tag)
             profile.tags.add(tag)
-    else:
-        _log.warning('No labels found in pod %s' % pod_name)
 
     _datastore_client.profile_update_tags(profile)
 
@@ -397,7 +398,7 @@ def _generate_rules(container_id, namespace, pod_name):
     :return two lists of rules(arg lists): inbound list of rules (arg lists)
     outbound list of rules (arg lists)
     """
-    ns_tag = _get_namespace_tag(namespace)
+    _log.info("Generating rules for pod %s", pod_name)
 
     # kube-system services need to be accessed by all namespaces
     if namespace == "kube-system" :
@@ -406,9 +407,12 @@ def _generate_rules(container_id, namespace, pod_name):
         return [["allow"]], [["allow"]]
 
     if namespace and DEFAULT_POLICY == 'ns_isolation':
+        _log.info("Creating rule to only allow traffic from namespace %s", namespace)
+        ns_tag = _get_namespace_tag(namespace)
         inbound_rules = [["allow", "from", "tag", ns_tag]]
         outbound_rules = [["allow"]]
     else:
+        _log.info("Creating rule to allow all incoming and outgoing traffic")
         inbound_rules = [["allow"]]
         outbound_rules = [["allow"]]
 
@@ -472,11 +476,11 @@ def _get_namespace_tag(namespace):
     """
     Pull metadata for namespace and return it and a generated NS tag
     """
-    ns_tag = _escape_chars('%s=%s' % ('namespace', ns))
+    ns_tag = _escape_chars('%s=%s' % ('namespace', namespace))
     return ns_tag
 
 
-def _escape_chars(self, unescaped_string):
+def _escape_chars(unescaped_string):
     """
     Calico can only handle 3 special chars, '_.-'
     This function uses regex sub to replace SCs with '_'
@@ -494,7 +498,12 @@ def _escape_chars(self, unescaped_string):
 def _get_metadata(namespace, pod_name, key):
     """Return the Metadata[key] of the pod to which the container belongs to"""
     pod_info = _get_pod_config(namespace, pod_name)
-    return pod_info["metadata"][key]
+    try:
+        metadata = pod_info["metadata"][key]
+    except KeyError:
+        _log.warning("Metadata for \"%s\" on pod \"%s\" not found.", key, pod_name)
+        metadata = None
+    return metadata
 
 
 def _get_pod_config(namespace, pod_name):
@@ -538,7 +547,7 @@ def _get_api_path(path):
     return json.loads(response_body)['items']
 
 
-def _get_api_token(self):
+def _get_api_token():
     """
     Get the kubelet Bearer token for this node, used for HTTPS auth.
     If no token exists, this method will return an empty string.
@@ -552,7 +561,7 @@ def _get_api_token(self):
             json_string = f.read()
     except IOError as e:
         _log.warning("Failed to open auth_file (%s). Assuming insecure mode", e)
-        if self._api_root_secure():
+        if _api_root_secure():
             _log.error("Cannot use insecure mode. API root is set to"
                          "secure (%s). Exiting", KUBE_API_ROOT)
             sys.exit(1)
@@ -564,7 +573,7 @@ def _get_api_token(self):
     return auth_data['BearerToken']
 
 
-def _api_root_secure(self):
+def _api_root_secure():
     """
     Checks whether the KUBE_API_ROOT is secure or insecure.
     If not an http or https address, exit.
--- a/calico_kubernetes/calico_kubernetes_cni.py
+++ b/calico_kubernetes/calico_kubernetes_cni.py
@@ -76,6 +76,7 @@ def create(args):
     interface = args['interface']
     net_name = args['name']
     subnet = args['subnet']
+    profile_name = '%s_%s_%s' % (namespace, pod_name, container_id[:12])
 
     _log.info('Configuring pod %s' % pod_name)
 
@@ -87,7 +88,7 @@ def create(args):
                              namespace=namespace,
                              endpoint=endpoint,
                              pod_name=pod_name,
-                             profile_name=net_name)
+                             profile_name=profile_name)
 
     dump = json.dumps(
         {
--- a/calico_kubernetes/calico_kubernetes_cni.py
+++ b/calico_kubernetes/calico_kubernetes_cni.py
@@ -253,40 +253,6 @@ def _set_profile_on_endpoint(container_id, namespace, endpoint, pod_name, profil
                                               endpoint_id=endpoint.endpoint_id)
 
 
-def _assign_default_rules(profile_name):
-    """
-    Generate a new profile rule list and update the _datastore_client
-    :param profile_name: The profile to update
-    :type profile_name: string
-    :return:
-    """
-    try:
-        profile = _datastore_client.get_profile(profile_name)
-    except:
-        _log.error("Could not apply rules. Profile not found: %s, exiting" % profile_name)
-        sys.exit(1)
-
-    rules_dict = {
-        "id": profile_name,
-        "inbound_rules": [
-            {
-                "action": "allow",
-            },
-        ],
-        "outbound_rules": [
-            {
-                "action": "allow",
-            },
-        ],
-    }
-
-    rules_json = json.dumps(rules_dict, indent=2)
-    profile_rules = Rules.from_json(rules_json)
-
-    _datastore_client.profile_update_rules(profile)
-    _log.info("Finished applying default rules.")
-
-
 def _assign_to_pool(subnet):
     """
     Take subnet (str), create IP pool in datastore if none exists.
@@ -652,7 +618,7 @@ def validate_args(env, conf):
     _log.debug('Environment: %s' % env)
     _log.debug('Config: %s' % conf)
 
-    args = dict()
+    args = {}
 
     # ENV
     try:
@@ -671,13 +637,19 @@ def validate_args(env, conf):
         sys.exit(1)
 
     try:
-        args['namespace'] = env['CNI_ARGS']['K8S_POD_NAMESPACE']
+        cni_args = dict(arg.split('=') for arg in env['CNI_ARGS'].split(';'))
+    except KeyError:
+        _log.error('No CNI_ARGS in Environment')
+        sys.exit(1)
+
+    try:
+        args['namespace'] = cni_args['K8S_POD_NAMESPACE']
     except KeyError:
         _log.error('No K8S_POD_NAMESPACE in Environment')
         sys.exit(1)
 
     try:
-        args['pod_name'] = env['CNI_ARGS']['K8S_POD_NAME']
+        args['pod_name'] = cni_args['K8S_POD_NAME']
     except KeyError:
         _log.error('No K8S_POD_NAME in Environment')
         sys.exit(1)
@@ -724,7 +696,6 @@ if __name__ == '__main__':
 
     pycalico_logger = logging.getLogger(pycalico.__name__)
     configure_logger(pycalico_logger, logging.DEBUG, False)
-
 
     # Environment
     env = os.environ.copy()
--- a/calico_kubernetes/calico_kubernetes_cni.py
+++ b/calico_kubernetes/calico_kubernetes_cni.py
@@ -17,6 +17,9 @@ import logging
 import json
 import os
 import sys
+import requests
+import re
+import sh
 
 from netaddr import IPAddress, IPNetwork, AddrFormatError
 from docker import Client
@@ -28,14 +31,21 @@ from pycalico.datastore_datatypes import Rules, IPPool
 from logutils import configure_logger
 
 ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
+CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/usr/bin/calicoctl')
+KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
+                               'http://kubernetes-master:8080/api/v1/')
 LOG_DIR = '/var/log/calico/kubernetes'
+
+DEFAULT_POLICY = os.environ.get('DEFAULT_POLICY', 'allow')
+POLICY_ANNOTATION_KEY = "projectcalico.org/policy"
 
 ORCHESTRATOR_ID = "docker"
 HOSTNAME = socket.gethostname()
 
 _log = logging.getLogger(__name__)
-datastore_client = IPAMClient()
-docker_client = Client()
+_datastore_client = IPAMClient()
+_docker_client = Client()
+_calicoctl = sh.Command(CALICOCTL_PATH).bake(_env=os.environ)
 
 
 def calico_kubernetes_cni(args):
@@ -60,18 +70,23 @@ def create(args):
     :param args: dict of values to pass to other functions (see: validate_args)
     """
     container_id = args['container_id']
+    namespace = args['namespace']
+    pod_name = args['pod_name']
     netns = args['netns']
     interface = args['interface']
     net_name = args['name']
     subnet = args['subnet']
 
-    _log.info('Configuring pod %s' % container_id)
+    _log.info('Configuring pod %s' % pod_name)
 
     endpoint = _create_calico_endpoint(container_id=container_id,
                                        interface=interface,
                                        subnet=subnet)
 
-    _set_profile_on_endpoint(endpoint=endpoint,
+    _set_profile_on_endpoint(container_id=container_id,
+                             namespace=namespace,
+                             endpoint=endpoint,
+                             pod_name=pod_name,
                              profile_name=net_name)
 
     dump = json.dumps(
@@ -83,7 +98,7 @@ def create(args):
     _log.info('Dumping info to kubernetes: %s' % dump)
     print(dump)
 
-    _log.info('Finished Creating pod %s' % container_id)
+    _log.info('Finished Creating pod %s' % pod_name)
 
 
 def delete(args):
@@ -103,11 +118,11 @@ def delete(args):
                       container_id=container_id)
 
     # Delete profile if only member
-    if datastore_client.profile_exists(net_name) and \
-       len(datastore_client.get_profile_members(net_name)) < 1:
+    if _datastore_client.profile_exists(net_name) and \
+       len(_datastore_client.get_profile_members(net_name)) < 1:
         try:
             _log.info("Profile %s has no members, removing from datastore" % net_name)
-            datastore_client.remove_profile(net_name)
+            _datastore_client.remove_profile(net_name)
         except:
             _log.error("Cannot remove profile %s: Profile cannot be found." % container_id)
             sys.exit(1)
@@ -126,7 +141,7 @@ def _create_calico_endpoint(container_id, interface, subnet):
     _log.info('Configuring Calico networking.')
 
     try:
-        _ = datastore_client.get_endpoint(hostname=HOSTNAME,
+        _ = _datastore_client.get_endpoint(hostname=HOSTNAME,
                                           orchestrator_id=ORCHESTRATOR_ID,
                                           workload_id=container_id)
     except KeyError:
@@ -158,19 +173,19 @@ def _container_add(hostname, orchestrator_id, container_id, interface, subnet):
     :param subnet (str): Subnet to allocate IP to
     :rtype Endpoint: Endpoint created
     """
-    # Allocate and Assign ip address through datastore_client
+    # Allocate and Assign ip address through _datastore_client
     pool, ip = _assign_to_pool(subnet)
 
     # Create Endpoint object
     try:
         _log.info("Creating endpoint with IP address %s for container %s",
                   ip, container_id)
-        ep = datastore_client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
+        ep = _datastore_client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
                                               container_id, [ip])
     except AddrFormatError:
         _log.error("This node is not configured for IPv%d. Unassigning IP "
                    "address %s then exiting." % ip.version, ip)
-        datastore_client.unassign_address(pool, ip)
+        _datastore_client.unassign_address(pool, ip)
         sys.exit(1)
 
     # Obtain the pid of the running container
@@ -180,7 +195,7 @@ def _container_add(hostname, orchestrator_id, container_id, interface, subnet):
     # set up the default routes.
     _log.info("Creating the veth with pid %s on interface %s", pid, interface)
     ep.mac = ep.provision_veth(PidNamespace(pid), interface)
-    datastore_client.set_endpoint(ep)
+    _datastore_client.set_endpoint(ep)
 
     return ep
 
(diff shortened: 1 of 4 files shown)


calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -28,14 +31,21 @@
 from logutils import configure_logger
 
 ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
+CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/usr/bin/calicoctl')
```
caseydavenport: The default value here should be "calicoctl" rather than "/usr/bin/calicoctl"

This means that sh.Command will search the current PATH for calicoctl.

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -227,50 +243,17 @@ def _set_profile_on_endpoint(endpoint, profile_name):
     """
     _log.info('Configuring Pod Profile: %s' % profile_name)
 
-    if not datastore_client.profile_exists(profile_name):
+    if not _datastore_client.profile_exists(profile_name):
         _log.info("Creating new profile %s." % (profile_name))
-        datastore_client.create_profile(profile_name)
-        # _assign_default_rules(profile_name)
+        _datastore_client.create_profile(profile_name)
+        _apply_rules(container_id, namespace, pod_name, profile_name)
```
caseydavenport: I think `_apply_rules` is not applicable any more after https://github.com/projectcalico/libcalico/pull/9, which is in the latest release of libcalico.

It might be worth including the code to update to libcalico v0.3.0 in this PR.

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -307,6 +290,297 @@ def _get_container_pid(container_id):
...
+    swap_char = '_'
+
+    # If swap_char is in string, double it.
+    unescaped_string = re.sub(swap_char, "%s%s" % (swap_char, swap_char), unescaped_string)
+
+    # Substitute all invalid chars.
+    return re.sub('[^a-zA-Z0-9\.\_\-]', swap_char, unescaped_string)
+
+
+def _get_metadata(namespace, pod_name, key):
```
caseydavenport: We want to minimize the number of times our plugin is hitting the API, since each access is quite costly.

We should be doing this once at startup (since we know we'll need the data) and storing it so that we can access our copy on demand without hitting the API again.

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -307,6 +290,297 @@ def _get_container_pid(container_id):
...
+             this_pod = pod
+             break
+         else:
+             raise KeyError('Pod %s not found', pod_name)
+     _log.debug('Got pod data %s', this_pod)
+     return this_pod
+
+
+def _get_api_path(path):
+    """Get a resource from the API specified API path.
```
caseydavenport: Let's make sure all of our docstrings are the same format.

```
def thing()
   """
   Text that
   uses multiple lines
   """
```
OR
```
def thing():
   """Text
   that uses multiple lines
   """
```

But not both.  I prefer the first one myself.

calico_kubernetes/tests/test_cni_plugin.py
```diff
@@ -16,6 +16,8 @@
 from mock import patch, Mock
 from netaddr import IPAddress, IPNetwork
 from pycalico.datastore_datatypes import IPPool
+import sys
```
caseydavenport: nooooooo :(

Fine for now, but this will all go away when we refactor the plugin to be object oriented.

calico_kubernetes/calico_kubernetes_cni.py
Symmetric: Pep 8 suggests a standard here: https://www.python.org/dev/peps/pep-0257/.

Option 1 above is ok (we do it in plenty of places), but 2 above is more nuanced. If you put text on the first line, it specifically should be the title summary of the function.

E.g. 
def foobar():
    """Foo the bar
    What's a foo? What's a bar? etc.
    """

I see fairly often folk doing
    """start my 
    line and continue it
    """
Which isn't going to render well in e.g. the Python help() command.

Symmetric: Also worth mentioning that using Pycharm's
```
@param
@type
```
docstring annotations is STRONGLY preferred; it lets you do static type checks in Pycharm that can catch a whole class of bugs.

L-MA: I will standardize doc strings when I do the OO restructuring.

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -60,19 +81,23 @@ def create(args):
     :param args: dict of values to pass to other functions (see: validate_args)
     """
     container_id = args['container_id']
+    namespace = args['namespace']
+    pod_name = args['pod_name']
     netns = args['netns']
     interface = args['interface']
     net_name = args['name']
-    subnet = args['subnet']
+    profile_name = '%s_%s_%s' % (namespace, pod_name, container_id[:12])
```
L-MA: I changed the profile name from net_name to this to make testing easier. Do we want to revert the profile name back to net_name?

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -60,19 +81,23 @@ def create(args):
...
 
-    _log.info('Configuring pod %s' % container_id)
+    _log.info('Configuring pod %s' % pod_name)
```
Symmetric: Should we keep the ID in this log too?

caseydavenport: Remember not to use % formatting in Python logs.

calico_kubernetes/calico_kubernetes_cni.py
Symmetric: Hmm, we don't want one profile per pod do we? It doesn't look like we have advanced policy in this plugin yet, is that right? And in advanced policy mode, how were we handling profiles?

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -103,30 +131,29 @@ def delete(args):
                       container_id=container_id)
 
     # Delete profile if only member
-    if datastore_client.profile_exists(net_name) and \
-       len(datastore_client.get_profile_members(net_name)) < 1:
+    if _datastore_client.profile_exists(profile_name) and \
```
Symmetric: This code isn't safe -- what happens if someone creates a pod somewhere else in the cluster that uses this profile?

Symmetric: If this is a known limitation we should at least doc it here and be clear about our plans to fix it (TODO: etc)

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -227,73 +252,93 @@ def _set_profile_on_endpoint(endpoint, profile_name):
...
-def _assign_default_rules(profile_name):
+def _assign_ip_address():
     """
-    Generate a new profile rule list and update the datastore_client
-    :param profile_name: The profile to update
-    :type profile_name: string
+    Assigns and returns an IPv4 address using the IPAM plugin specified in CONFIG.
     :return:
     """
+    # May throw CalledProcessError - let it.  We may want to replace this with our own Exception.
```
Symmetric: I'd prefer we re-raise a custom exception; good practice to encapsulate error handling as close as possible to the code that's producing it. Any callers of _assign_ip_address shouldn't need to know (or care) that it throws a CalledProcessError.

calico_kubernetes/calico_kubernetes_cni.py
L-MA: This PR does include advanced policy - it pulls fine grained policy rules from the pod's annotations. In the old plugin we were doing one profile per pod.

calico_kubernetes/calico_kubernetes_cni.py
L-MA: This can be fixed if we agree to use one profile per pod.

Symmetric:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 224 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_OTT07qKCPIb2koCL)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L224)):</sup>
Paranoia: if _unassign_ip_address fails for an unforeseen reason, we'll dealloc the IP without deleting the endpoint. I'd prefer to delete the endpoint first, and have the consequence of an unforeseen failure be leaked IPs.

I think that concern might go away if we Pokemon catch in _unassign_ip_address function, but is there a reason to depend on that?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

calico_kubernetes/calico_kubernetes_cni.py
L-MA: In that case, shouldn't we be catching the exception in the _call_ipam_plugin function?

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -193,32 +220,30 @@ def _container_remove(hostname, orchestrator_id, container_id):
...
-    datastore_client.remove_workload(hostname=hostname,
+    _datastore_client.remove_workload(hostname=hostname,
                                      orchestrator_id=orchestrator_id,
                                      workload_id=container_id)
 
     _log.info("Removed Calico interface from %s" % container_id)
 
 
-def _set_profile_on_endpoint(endpoint, profile_name):
+def _set_profile_on_endpoint(container_id, namespace, endpoint, pod_name, profile_name):
```
L-MA: I dont really like the name of this function since we're doing a lot more than just setting the profile on endpoint. How about `_configure_profile`?

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -103,30 +131,29 @@ def delete(args):
...
-            datastore_client.remove_profile(net_name)
+            _log.info("Profile %s has no members, removing from datastore" % profile_name)
+            _datastore_client.remove_profile(profile_name)
         except:
             _log.error("Cannot remove profile %s: Profile cannot be found." % container_id)
             sys.exit(1)
 
 
-def _create_calico_endpoint(container_id, interface, subnet):
+def _create_calico_endpoint(container_id, interface):
```
L-MA: Is this wrapper necessary? Seems better to just move the endpoint check into `_container_add` and removing this function.

Symmetric:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 134 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_MLYH_9PjrWoh9_AK-r9-134)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L134)):</sup>
@caseydavenport,  have you thought about this one at all?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

caseydavenport:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 134 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_MLYH_9PjrWoh9_AK-r9-134)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L134)):</sup>
I actually wasn't aware we did this - I thought we removed this code a while back.  As it is, it shouldn't be a problem since we always have a profile per pod, we won't hit the race.  

However, this will need to change when we look at sharing profiles for better perf.  I don't think we can get away with a compare-and-delete operation in etcd, since the profile object doesn't change I don't think when we add endpoints to it.  Might require locking or the new data model that some people have been tossing about.  Definitely needs more thinking.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

calico_kubernetes/calico_kubernetes_cni.py
```diff
@@ -307,6 +352,337 @@ def _get_container_pid(container_id):
...
+
+    if annotations and POLICY_ANNOTATION_KEY in annotations:
+        _log.info("Getting policy rules from annotation of pod %s", pod_name)
+
+        # Remove Default Rule (Allow Namespace)
+        rules = annotations[POLICY_ANNOTATION_KEY]
+
+        # Rules separated by semicolons
+        for rule in rules.split(";"):
+            args = rule.split(" ")
```
Symmetric: I think it's cleaner to say `args = rule.split()`. The default split will turn `a  b` into ['a', 'b'], so you don't have to do your filter() below (and it'll handle other whitespace too, like tabs). That should prevent straggling empty list members, both inside your list and at the ends.

Symmetric:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 588 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tFgefkPISfBDgGpxm)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L588)):</sup>
This will lose information about the escaped characters, so we can't unescape them. That means that we will have collisions on policies (e.g. `tag*1` == `tag+1`). The original design converted to hex character codes; did we change that deliberately?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

caseydavenport:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 588 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tFgefkPISfBDgGpxm)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L588)):</sup>
Yes - this was deliberate.  AP was against properly escaping these since the resulting string was less human-readable, and real-world clashes seemed unlikely.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

Symmetric:
Reviewed 1 of 3 files at r8, 3 of 3 files at r9.
Review status: all files reviewed at latest revision, 36 unresolved discussions, all commit checks successful.

---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 102 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t3SZkomVpLDpnPWRU)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L102)):</sup>
`dump` isn't a very descriptive variable name; how about `json_result` or `json_output`?


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 111 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t0qijLIZwTUw0z77H)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L111)):</sup>
Don't use % in log statements (pass variables as args to the log function instead).


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 144 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_MLYH_9PjrWoh9_AK-r9-144)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L144)):</sup>
It's borderline. Functions are clearer to read if all of the operations are at the same level; the create() function is a composition of a few high-ish level functions. Also, functions should strive to be shorter than a screen's length; easier to test small functions that do one thing than god functions that do everything.

I'm inclined to keep the current split, because if create() was any bigger then inlining this function would make it too big.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 190 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t3wl7sOe5hBCoZ4LY)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L190)):</sup>
We should log e.output and e.returncode, in case the command gave us something useful.

Also, `except Exception, e:` is a deprecated syntax; it's ambiguous, so you should always use `except Exception as e:`.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 191 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t4ZW9yG4QVe3oBp9a)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L191)):</sup>
I'm not sure it makes sense to return the underlying error code here; the caller has no idea what those codes correspond to. Either we should return 1 on all failures, or we should synthesize our own error codes (i.e. 1=general failure, 2=ipam failure, etc). I'd go with the former for now.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 201 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t4sBWdUEMVkNq77_m)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L201)):</sup>
This should probably be a constant.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 246 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_MLYH_9PjrWoh9_AK-r9-246)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L246)):</sup>
We used to have a _configure_profile; I think that better describes what's going on.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 261 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t5FaEs65c0Nlxphfc)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L261)):</sup>
Can we make a better error message than "Cannot create profile"? Are we assuming that a better error was logged inside create_profile? This might be OK, but we should leave a comment explaining why, if that's the case. But ideally we can beef up the error here (e.g. catch an EtcdException and write a descriptive log; fall back to this error to catch unexpected errors).


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 277 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_MLYH_9PjrWoh9_AK-r9-277)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L277)):</sup>
Yep, that's the right place. Should docstring the expected exceptions there: `""" throws CalicoIPAMError if ..."""` or something like that.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 283 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t6GKdhsZMwN72P7bd)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L283)):</sup>
I'd prefer you use a new variable `json_result` or something like that; reassigning an existing variable can be OK, but it can also mask programming errors. I generally would reuse the same variable if it's the same type (e.g. you're stripping whitespace on string input), but not when you're transforming to a different type as we are here.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 333 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-_QTazkrmABOBXf1rv)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L333)):</sup>
Popen is almost never the right answer. Why not check_output()?


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 346 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t70yvZcbKDIiP2jr4)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L346)):</sup>
Indenting looks funny here. Too many spaces? Also suggests that we're not covering this line with UTs... can you fix that too if it's the case?


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 384 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t7nlHLpn4Fp2JZ2LK)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L384)):</sup>
I think this comment should be fleshed out a bit -- why are we assuming this? Critically, what bit of code/behaviour do I need to check to see if that's still the case?


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 445 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t8AmHEaaNjnc24bCY)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L445)):</sup>
"Adding namespace tag...", to disambiguate from the "Adding tag" log below. (Rule of thumb -- never make the same log in two different places; it makes it harder to work backwards from a log to the code).

Also, `s/ %/,/`


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 455 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t8XJLgPqvIT9pDYKx)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L455)):</sup>
`s/ +/,/` (don't do string interpolation unless needed).


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 485 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-t9FW6dV1jZ7Q7MWPd)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L485)):</sup>
Is this what we want? If we're doing ns_isolation, but `namespace==None`, then we'll fall through to the `allow all` branch, which I don't think is correct.

I think we should nest:

```
if POLICY==ns_isolation:
  if ns==kube-system:
    allow all
  else:
    assert namespace
    allow namespace
else:
  allow all
```

Alternatively could de-dupe, like so:
```
if POLICY==ns_isolation && ns != kube-system:
  assert namespace
  allow namespace
else:
  allow all
```

This is less code, but the logic is slightly harder to grok; I don't mind duplicating two lines (approach 1) to make the flow clearer.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 506 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tBe5ac_lqg-ncUJU0)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L506)):</sup>
Typo -- can you run through PyCharm's warning list to catch errors like this? Ideally you'd have no warnings in your IDE.

For things like PyCharm's "Too broad exception" error, OK to add pycharm-specific annotations like "        # noinspection PyBroadException" (Though always add your own comment describing why it's ok to disable the inspection).


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 534 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tCkgBnX5zu2zfZZVH)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L534)):</sup>
This code is brittle; if there is no "=", then we'll hit an error here (since that will turn into "key,value = ["label"]. If there are two "=" characters hilarity will also ensue.

This is a job for regexes; we should specify a strict match on `label (\w+)\=(\w+) ` (note trailing whitespace), and log an error if the regex doesn't match. Open question as to whether we should allow a pod to start if one of its rules failed to parse; the principle of least surprise suggests that we should fail fast here, rather than programming only some of the rules that were requested.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 597 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tIYTNRdFi7De_iQTR)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L597)):</sup>
Is this really a warning-level log? Probably most pods won't have the 'label' or 'annotations' metadata keys, so I suspect it's not exceptional. (I think we shouldn't log at all, and just log in the caller if it's unexpected for this to reply with no keys).


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 598 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tImif_opC7YmDaE4f)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L598)):</sup>
Should be `metadata = []` for consistency, since we return a list in the success case. Then callers can iterate over the returned list safely (rather than having to test for None, just do `for k in metadata`, which is a no-op for an empty list. If having no metadata is exceptional, the caller can say `if not metadata: <log exception>`.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 615 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tJmMcC3QPKgficWq2)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L615)):</sup>
This indentation is not correct -- want the else to be branched from the for:, not the if:.

Also, the indentation in this function is not correct (5 chars not 4?), as discussed.


---

<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 780 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tN4wzwWVIZb7XOnB7)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L780)):</sup>
What's this for? Looks like you're also checking that the ipam block has a type, but that should give a different error if that's what you're checking. 

Also, shouldn't that cause an error in the ipam module itself? Do we really care what's in the ipam blob?

A comment would be useful to explain what's going on, if you do keep this check.


---

<sup>**[calico_kubernetes/tests/test_cni_plugin.py, line 183 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-te9JodgtgXP4ECGn_)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/tests/test_cni_plugin.py#L183)):</sup>
We should be testing the failure cases for the IPAM plugin too, both at the UT level (i.e. calling function handles failure properly) and the FV level (call the top-level create(), stub out non-plugin calls, and check that we get the right output.)

Probably fair to leave FVs as a TODO: in the code for now, but we should get them on the agenda. Definitely need to UT failure cases before we can merge this PR though.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

Symmetric:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 588 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tFgefkPISfBDgGpxm)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L588)):</sup>
At very least we should make a comment explaining this. And the design doc should probably be mothballed (since it doesn't reflect reality any more). I don't think this is a production-ready design, but it's probably OK for our current demo implementation.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

Symmetric:
The plugin should probably still be called 'calico_kubernetes.py', assuming we're not planning to keep the two side-by-side.

caseydavenport:
<sup>**[calico_kubernetes/calico_kubernetes_cni.py, line 588 \[r9\]](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53#-K-tFgefkPISfBDgGpxm)** ([raw file](https://github.com/projectcalico/calico-kubernetes/blob/aa838646344af1b76309287195a60ee1b6bec528/calico_kubernetes/calico_kubernetes_cni.py#L588)):</sup>
Yep, the design doc can go away at this point.

FYI we are tracking this: https://github.com/projectcalico/calico-kubernetes/issues/8


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/projectcalico/calico-kubernetes/53)

caseydavenport:
Closing for now since this is quite old.

PR #56, opened 2015-09-30
Log out exceptions thrown in the plugin

### LINKED ISSUE #46: Plugin Errors not emitted to log.
I see the plugin crashing (due to a configuration error) but it is not emitted in the logfile.

caseydavenport:
Fixes #46

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,8 +60,5 @@ target/
 # Emacs temp
 *~
 
-# Pycharm stuff
-.idea/
-
 # Build markers
 *.created
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ target/
 
 # Pycharm stuff
 .idea/
+
+# Build markers
+*.created
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -710,10 +710,11 @@ class NetworkPlugin(object):
         return tag
 
 
-if __name__ == '__main__':
-    configure_logger(logger, LOG_LEVEL, True)
-    configure_logger(pycalico_logger, LOG_LEVEL, False)
-
+def main():
+    """
+    Runs the plugin.
+    """
+    logger.info('Plugin Args: %s', sys.argv)
     mode = sys.argv[1]
 
     if mode == 'init':
@@ -724,7 +725,6 @@ if __name__ == '__main__':
         pod_name = sys.argv[3].replace('/', '_')
         docker_id = sys.argv[4]
 
-        logger.info('Args: %s', sys.argv)
         logger.info("Using LOG_LEVEL=%s", LOG_LEVEL)
         logger.info("Using ETCD_AUTHORITY=%s", os.environ[ETCD_AUTHORITY_ENV])
         logger.info("Using CALICOCTL_PATH=%s", CALICOCTL_PATH)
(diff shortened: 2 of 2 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -710,10 +710,11 @@ def _label_to_tag(self, label_key, label_value):
...
 
-if __name__ == '__main__':
-    configure_logger(logger, LOG_LEVEL, True)
-    configure_logger(pycalico_logger, LOG_LEVEL, False)
-
+def main():
+    """
+    Runs the plugin.
+    """
+    logger.info('Plugin Args: %s' % sys.argv)
```
alexwlchan: Better to use

```python
logger.info('Plugin Args: %s', sys.argv)
```

This variant only performs the string substitution if the log is above the log-level threshold; otherwise it gets skipped.

caseydavenport: Yep, good catch.  That seems to be the only one of the logs that was done that way....

caseydavenport: And apparently a bunch of others scattered throughout.... Saw your PR.  Thanks ;)

.gitignore
```diff
@@ -61,4 +61,7 @@ target/
 *~
 
 # Pycharm stuff
-.idea/
\ No newline at end of file
+.idea/
```
Symmetric: This should really go in your ~/.gitignore file BTW, but I know it was here already.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -741,3 +741,23 @@ def _label_to_tag(self, label_key, label_value):
...
+    configure_logger(pycalico_logger, LOG_LEVEL, False)
+
+    # Try to run the plugin, logging out any BaseExceptions raised.
+    logger.info("Begin Calico network plugin execution")
+    try:
+        main()
+    except SystemExit:
+        # If a SystemExit is thrown, we've already handled the error and have called sys.exit()
+        # No need to produce a duplicate exception message, so exit without logging.
+        os._exit(1)
```
Symmetric: Why catch SystemExit and then os._exit? Why not just os._exit inside the call?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -741,3 +741,23 @@ def _label_to_tag(self, label_key, label_value):
...
+    except BaseException:
+        logger.exception("Unhandled Exception killed plugin")
+        os._exit(1)
+    finally:
```
Symmetric: This isn't going to be hit if you os._exit()

calico_kubernetes/calico_kubernetes.py
Symmetric: I think we should actually not os._exit here, just pass -- and finally: sys.exit(). That way we cleanup correctly.

caseydavenport: >Why catch SystemExit and then os._exit? Why not just os._exit inside the call?

This allows us to continue to use sys.exit() in our code, which is a lot easier to UT using `self.assertRaises(some_func, SystemExit)`.

Symmetric: Makes sense -- I still think my second point stands, since os._exit is actually potentially dangerous, as it doesn't clean up file handles etc.

caseydavenport: Yep, agreed.  I'll make that change.

caseydavenport:
@Symmetric - I think I've addressed all your comments. Care to give another look?

Symmetric:
LGTM

PR #58, opened 2015-10-02
Handle duplicate IP assignments when using Docker IPAM

caseydavenport:
Seen when restarting docker service or Kubernetes node.

Partial workaround for kubernetes/kubernetes#14940

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -10,7 +10,7 @@ from subprocess import check_output, CalledProcessError, check_call
 import requests
 import sh
 import logging
-from netaddr import IPAddress, AddrFormatError
+from netaddr import IPAddress, IPNetwork, AddrFormatError
 from logutils import configure_logger
 import pycalico
 from pycalico import netns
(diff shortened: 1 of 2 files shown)


caseydavenport:
Working on tests for this now, but otherwise should be good to review.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -265,28 +265,58 @@ def _assign_container_ip(self):
 
         :return IPAddress which has been assigned
         """
+        def _assign(ip):
```
Symmetric: Why have you made this a nested function, instead of a method in its own right? Reckon the call stack would be cleaner in the latter formulation.

caseydavenport: Yeah, good point.  I likely this because you didn't end up with multiple `assign` methods like `_assign_container_ip` and `__actually_assign_ip` or something like that, but I can probably make something work.

Symmetric:
One structuring question, aside from that looks good.

calico_kubernetes/calico_kubernetes.py
Symmetric: Cool, not a massive issue so I won't block this going in if you can't come up with something cleaner.

PR #60, opened 2015-10-05
Log INFO and above to stdout

caseydavenport:
Also includes a refactor to keep line lengths less than 79 characters.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -70,7 +70,9 @@ class NetworkPlugin(object):
         self.pod_name = pod_name
         self.docker_id = docker_id
         self.namespace = namespace
-        self.profile_name = "%s_%s_%s" % (self.namespace, self.pod_name, str(self.docker_id)[:12])
+        self.profile_name = "%s_%s_%s" % (self.namespace,
+                                          self.pod_name,
+                                          str(self.docker_id)[:12])
 
         logger.info('Configuring docker container %s', self.docker_id)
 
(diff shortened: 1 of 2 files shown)


caseydavenport:
@Symmetric - are you interested in reviewing this?

Symmetric:
Yup, will look now.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -725,22 +755,26 @@ def main():
...
-            logger.info('Executing Calico pod-deletion hook')
-            NetworkPlugin().delete(namespace, pod_name, docker_id)
-        elif mode == 'status':
+        if mode == "status":
+            # Status is called on a regular basis - handle separately
+            # to avoid flooding the logs.
             logger.info('Executing Calico pod-status hook')
             NetworkPlugin().status(namespace, pod_name, docker_id)
+        else:
+            logger.info("Using LOG_LEVEL=%s", LOG_LEVEL)
```
Symmetric: Surely these are debug logs?

Symmetric:
LGTM

calico_kubernetes/calico_kubernetes.py
caseydavenport: I'd kind of like to print them as info logs during `init` and then as debug logs other time.

On that note, it might be worthwhile to do some validation in `init` - e.g check we can reach the KUBE_API_ROOT and ETCD_AUTHORITY.  Might make debugging easier, since the kubelet will fail to start if the plugin is misconfigured (rather than starting but failing the create pods).  Just a thought.

Symmetric: Worth mentioning that INIT is going away in CNI, so moving stuff over there would be temporary.

PR #61, opened 2015-10-05
Add coveralls support

Symmetric:
Add the coveralls package to the requirements.txt. After
applying our repo key to the CircleCI config, this should
enable Coveralls reports on CircleCI.


caseydavenport:
Nice, I was thinking of doing this myself at some point as well.

LGTM.

Issue #11, opened 2015-08-12
Stop using `calicoctl` for profile configuration


alexhersh:
Currently, semantic rules can only be added by `calicoctl` one at a time. Ideally we want to support profile _updating_ as opposed to _appending_ with this format. We also do not want this plugin to call `calicoctl` at all, but some work needs to be done to allow the pycalico datastore to accept our semantic format. 

A symptom of this is that we need to flush out the default rules before adding in new policy. This is currently done naively with `calicoctl` calls that target rules that are assumed to exist. If these defaults change, new policy will be disrupted. Default rule replacement should be more dynamic (and preferably without `calicoctl` calls)

caseydavenport:
Changing to low priority, since ideally this will just go away once we can get proper policy support in the Kubernetes API (either native, pluggable, etc) and we no longer need to use the calicoctl styled policy statements in annotations.

It would be nice to remove the calicoctl dependency once and for all, but I don't think we should move the argparsing stuff into libcalico, especially if its just for this use case.

caseydavenport:
The last remaining calls to `calicoctl`are applying advanced policy rules.  To remove them, we need to write some parsing code to interpret human-readable policy rules and convert them to `libcalico` Rule objects.

There might be some common function here between what the `calico-kubernetes` plugin does, and what `calicoctl` does.  However, I think that the two are logically separate enough that it makes sense to keep the parsing code separate (e.g the rules that the calico-kubernetes plugin interprets are not exactly the same as the calicoctl rules and may diverge further down the road).

caseydavenport:
Fixed by #64

PR #64, opened 2015-10-06
Parse policy statements without `calicoctl`

caseydavenport:
Addresses #11


caseydavenport:
@Symmetric - No rush to get started on this (since I'm out for the rest of the week and can't address comments).

calico_kubernetes/calico_kubernetes.py
```diff
@@ -56,20 +65,19 @@ def __init__(self):
...
             version=DOCKER_VERSION,
             base_url=os.getenv("DOCKER_HOST", "unix://var/run/docker.sock"))
 
     def create(self, namespace, pod_name, docker_id):
         """"Create a pod."""
-        # Calicoctl does not support the '-' character in iptables rule names.
-        # TODO: fix Felix to support '-' characters.
         self.pod_name = pod_name
         self.docker_id = docker_id
         self.namespace = namespace
```
Symmetric: Can we rename this `k8s_namespace` or something like that? `namespace` is overloaded with `network_namespace`.

caseydavenport: I'm fine with doing that.  However, we don't have a concept of "Network Namespace" in this code currently, so I wouldn't prioritize this immediately.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -568,120 +574,53 @@ def _api_root_secure(self):
...
+                # Allow all traffic by default.
+                logger.debug("Default policy is allow all")
+                inbound_rules = [allow]
+                outbound_rules = [allow]
+
+            logger.info("Checking for annotation based policy")
+            annotations = self._get_metadata(pod, "annotations")
+
+            # If policy annotations are defined, use them to generate rules.
+            if annotations and POLICY_ANNOTATION_KEY in annotations:
```
Symmetric: Might be clearer if we make this part of the if/elif branch above? That way we only set the inbound_rules once, and don't rely on overwriting a previously-set value.

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,343 @@
...
+        Takes a single line of policy as defined in the annotations of a
+        pod and returns the equivalent libcalico Rule object.
+        :param policy: Policy string from annotations.
+        :return: A Rule object which represent the given policy.
+        """
+        _log.info("Parsing policy line: '%s'", policy)
+        splits = self.validate_policy(policy)
+    
+        # Parse based on the protocol.
+        if "tcp" in splits:
```
Symmetric: I hate to drop this feedback on a fully-formed PR... but I am not sure about this approach. I think we should be using a proper parsing library to build up an AST for our rules.

The cheapo implementation that I was imagining was to use docopt to build this (we do have a working docopt parser for the command already, it's in the calicoctl Usage string). That way we guarantee that the command will be parsed in the same way as calicoctl would (assuming that the docstrings don't diverge).

The less cheap way would be to use [Pyparsing](https://pyparsing.wikispaces.com/), which lets you build arbitrary parsers in a clean pythonic way. Then you can do something like

```python
parser = Optional(protocol) + Optional(from_clause) + Optional(to_clause)
```

Your approach clearly works for the test cases that you've written, and it may well be correct, but it's much less obviously correct than the parser-based approaches above, which would probably be tens of lines of code on our side.

Thoughts?

caseydavenport: Yep you're right, was probably a bit silly of me to jump the gun on this one..  When I got started I thought my approach was going to be a bit cheaper than it turned out to be...

I'll take a look at library options here (inc Pyparsing and docopt).  My first impression is that docopt is not the right tool to use here, and I'd rather use something homegrown, but Pyparsing looks interesting and it might be worth swapping over to it.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Done.

calico_kubernetes/policy.py
caseydavenport: Done - refactored to use docopt

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,224 @@
...
+        if arguments.get("tcp"):
+            protocol = "tcp"
+        elif arguments.get("udp"):
+            protocol = "udp"
+        elif arguments.get("icmp"):
+            protocol = "icmp"
+        else:
+            protocol = None
+
+        src_ports = arguments.get("<SRCPORTS>")
```
Symmetric: Can you do away with these intermediate variables by referencing arguments directly in the _validate_X calls?

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,224 @@
...
+        if src_label:
+            rule_args["src_tag"] = self._validate_label(src_label)
+        if dst_net:
+            rule_args["dst_net"] = self._validate_ip(dst_net)
+        if dst_label:
+            rule_args["dst_tag"] = self._validate_label(dst_label)
+
+        return Rule(**rule_args)
+
+    def _validate_int(self, value):
```
Symmetric: Why are we validating these inputs here? Shouldn't the downstream code throw an exception if it's not valid? (Better to ask forgiveness etc.)

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        if not "=" in value:
+            raise PolicyParseError("Failed to parse %s, expecting "
+                                   "label of form X=Y", value)
+        k,v = value.split("=")
+        tag = self._label_to_tag(k, v)
+        return tag
+    
+    
+    def _validate_ip(self, value):
```
Symmetric: Again, downstream code surely must fail if we don't pass a valid IP?

calico_kubernetes/policy.py
caseydavenport: Yep.

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,224 @@
...
+            raise PolicyParseError("Failed to parse %s, expecting int", value)
+
+    def _validate_label(self, value):
+        """
+        Takes the given label, validates it, and returns the equivalent tag.
+        """
+        if not "=" in value:
+            raise PolicyParseError("Failed to parse %s, expecting "
+                                   "label of form X=Y", value)
+        k,v = value.split("=")
```
Symmetric: This isn't going to render `a=b=c` correctly. Can we insert this into the docopt parse string? If you can't do arbitrary k=v there, then I'd prefer you did a regex here so that we know there's only one `=`.

calico_kubernetes/policy.py
caseydavenport: Unfortunately, I don't think that is the case. libcalico doesn't validate these things, and we'll only learn about errors when felix hits them (which will be hard to diagnose, and too late for the plugin to respond)

calico_kubernetes/policy.py
```diff
@@ -0,0 +1,224 @@
...
+
+        return rule
+
+    def _generate_rule(self, arguments):
+        """
+        Generates a libcalico
+        :param arguments: A dictionary of arguments with the following format.
+
+        {
+            "<DSTCIDR>": null,
```
Symmetric: Do we need to reproduce this here, since the docstring is the canonical representation?

calico_kubernetes/policy.py
Symmetric: I'd say there's a good argument for adding asserts in libcalico, rather than doing the arg checking in the caller. That way other clients don't have to remember to do this error-checking, nor do they need to reimplement it.

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -416,16 +416,16 @@ def test_delete_docker_interface(self):
...
-                          autospec=True) as m_datastore_client, \
-                patch_object(self.plugin, '_get_namespace_tag',
-                             autospec=True) as m_get_namespace_tag, \
-                patch_object(self.plugin, '_get_pod_config',
-                             autospec=True) as m_get_pod_config, \
-                patch_object(self.plugin, '_apply_rules',
-                             autospec=True) as m_apply_rules, \
-                patch_object(self.plugin, '_apply_tags',
-                             autospec=True) as m_apply_tags:
+        with patch.object(self.plugin, '_datastore_client',
```
Symmetric: As I pointed out to Alex -- you don't need a patch.object here, since you're creating a new plugin for each test case. Just do `m_datastore_client = MagicMock(); self.plugin._datastore_client = m_datastore_client`.

I don't think you should bother changing this throughout the file, but worth fixing up wherever you're making edits, since it reduces the indentation (and gains us precious characters on each line)

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -576,50 +579,78 @@ def test_get_api_path(self, m_json_load, m_session):
...
+                         outbound_rules=outbound)
+
+        # Call method under test empty annotations/namespace
+        return_val = self.plugin._generate_rules(pod)
+
+        # Assert
+        self.assertEqual(return_val, expected)
+
+    def test_generate_rules_annotations(self):
+        pod = {'metadata':
```
Symmetric: Weird formatting

calico_kubernetes/policy.py
caseydavenport: We don't need to include this here.  However, strictly speaking, there is nothing that states the argument to this method is related in any way to the output of "docopt" (though in practice we know it currently is).

Symmetric: I'd be more inclined to make this function's docstring contain the docopt usage (if possible), or make the docopt usage string a variable inside this function, rather than duplicate the text.

calico_kubernetes/policy.py
caseydavenport: I think the way it is now is cleaner - as we nest the arguments.get(...) calls it makes lines longer and harder to read.  I'd prefer to leave it this way.

calico_kubernetes/policy.py
caseydavenport: I'd also say that.  Shall I remove this and open a new PR for libcalico?  That would gate this PR on the libcalico changes / release.

calico_kubernetes/tests/policy_test.py
```diff
@@ -0,0 +1,183 @@
...
+        self.assertEqual(parsed['src_tag'], expected_tag)
+
+    def test_parse_allow(self):
+        rule = "allow"
+
+        # Parse the rule
+        parsed = self.parser.parse_line(rule)
+
+        # Assert on the result
+        self.assertEqual(parsed['action'], "allow")
```
Symmetric: Worth throwing in a negative test or two, to check that we don't have an icmp_port, dst_net, etc? I don't think you have to do that on every TC, but would be good to sanity-check at least once.

calico_kubernetes/tests/policy_test.py
```diff
@@ -0,0 +1,183 @@
...
+        self.assertEqual(parsed['protocol'], "icmp")
+        self.assertEqual(parsed['icmp_type'], 8)
+        self.assertEqual(parsed['src_net'], IPNetwork("1.2.0.0/16"))
+
+    def test_parse_errors(self):
+        # Missing icmp type.
+        rule = "allow icmp type code 4"
+        self.assertRaises(PolicyParseError, self.parser.parse_line, rule)
```
Symmetric: This code would be more concise IMO if you used the @parameterized decorator. You'd trim n spaces, and n assertRaises.

calico_kubernetes/tests/policy_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Invalid icmp type.
+        rule = "allow icmp type ten code 4"
+        self.assertRaises(PolicyParseError, self.parser.parse_line, rule)
```
Symmetric: Prefer assert_raises from nosetools, since it gives pep8 compliant versions of the UnitTest functions.

Symmetric:
Overall looks good, a few nits on small-scale structure and test syntax.

calico_kubernetes/policy.py
caseydavenport: We're not really duplicating any text - the docopt usage string is quite different from this dictionary of arguments, to the extent that it is hard to guess what this dictionary will look like given the docopt usage string.

I'm happy to remove this from the docstring if you think it is overkill, but this isn't the place for the docopt usage string to live - docopt has been called and is out of the picture at this point in the execution.

calico_kubernetes/policy.py
Symmetric: Probably not worth holding this PR back for that I suppose; should be a cheap PR so maybe TODO' the change here?

calico_kubernetes/policy.py
caseydavenport: Yep, good spot.  Switched to regex.

calico_kubernetes/policy.py
caseydavenport: Yep, done.

calico_kubernetes/policy.py
caseydavenport: Added TODO

calico_kubernetes/tests/kube_plugin_test.py
caseydavenport: Done.

calico_kubernetes/tests/policy_test.py
caseydavenport: Fixed.

calico_kubernetes/tests/kube_plugin_test.py
caseydavenport: Fixed.

calico_kubernetes/tests/policy_test.py
caseydavenport: Fixed.

calico_kubernetes/tests/policy_test.py
caseydavenport: fixed.

calico_kubernetes/policy.py
Symmetric: I'm ok either way, up to you.

calico_kubernetes/policy.py
caseydavenport: I've also raised projectcalico/libcalico#37 to track this.

caseydavenport:
@Symmetric - I think I've addressed your outstanding comments.  Back to you!

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -416,42 +416,45 @@ def test_delete_docker_interface(self):
...
-            pod_name = 'pod_name'
-            profile_name = 'profile_name'
-            self.plugin.pod_name = pod_name
-            self.plugin.profile_name = profile_name
-
-            # Set up args
-            endpoint = Endpoint(TEST_HOST, TEST_ORCH_ID, '1234', '5678',
-                                'active', 'mac')
+        # Set up mock objects
+        m_datastore_client = Mock()
```
Symmetric: These should be of the form `MagicMock(spec=DatastoreClient)` (I think you can also do `MagicMock(spec=self.plugin._datastore_client)` if you think that's clearer. The latter might be more robust to code changes in the mocked class).

Symmetric: After thinking more about this suggestion (Alex also didn't spot the need to do `spec=`), I am on the fence as to whether this new way is safer than just using patch.object as a decorator with `autospec=True`.

I think it's about the same number of characters, so I'll leave it up to your preference. I do like the saving of indentation though, so the other option would be to use patch.object as a decorator instead of a context manager.

Symmetric:
Just one comment about Mock spec'ing, otherwise good to go.

calico_kubernetes/tests/kube_plugin_test.py
caseydavenport: Good spot.  We can't use `patch.object` as a decorator on `self.plugin` since `self` is not accessible in the decorator.  I think using `patch.object` is marginally safer, but you need to remember to spec anyway (either autospec, or explicitly) so I think I favor using MagicMock() to save the indentation.

caseydavenport: The datastore client should be mocked for all tests - this can be done in `setUp` to remove a lot of duplication.

Symmetric: I think Evan did this in his [coverage PR](https://github.com/projectcalico/calico-kubernetes/pull/71)

caseydavenport:
Rebased to master and fixed the Mock spec'ing issue.  Once the tests pass, I'll merge.

PR #65, opened 2015-10-06
Add logging of namespace state to plugin

Symmetric:
Previously we were not logging the state of the pod infra
container's namespace during our modifications, which make it
impossible to diagnose issues with these steps.

Add some verbose debug logging to show the interface state
on the machine before and after we configure the pod namespace.


caseydavenport:
@Symmetric - Tests are failing

Symmetric:
@caseydavenport fixed, PTAL

caseydavenport:
LGTM.

Issue #14, opened 2015-08-17
Support different types of authentication.


caseydavenport:
Support the various types of authentication for plugins as described here.

https://github.com/kubernetes/kubernetes/blob/master/docs/admin/authentication.md

Symmetric:
Initial design thinking: I don't think we need to support all of these auth methods; as long as we have one easy-to-configure way of granting Calico access.

My current plan is to implement bearer token support (probably replacing basic auth), and include instructions for either using Kubelet's token (if that's what kubelet is using), or generating a read-only ServiceAccount token for Calico.

PR #69, opened 2015-10-08
Service account auth

Symmetric:
Previously we extracted the Bearer token (used for authenticating
with the kube-apiserver) from the basic auth file. This auth
method is deprecated, so we should instead support the
more generic Service Account tokens.

Both mechanisms use the HTTP "Authentication: Bearer" header,
so all this commit changes is where we get that token from; now
we use the AUTH_TOKEN environment variable.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -56,6 +56,7 @@ class NetworkPlugin(object):
         self.profile_name = None
         self.namespace = None
         self.docker_id = None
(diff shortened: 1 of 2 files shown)


Symmetric:
Fixes #14.

caseydavenport:
Is there a corresponding PR for docs updates?

Symmetric:
@caseydavenport yes, https://github.com/alexhersh/calico-docker/commit/45ef883791a37a9b1eb542b869a522929b7c6c24

caseydavenport:
LGTM

Issue #67, opened 2015-10-06
We should have 100% code coverage


caseydavenport:
At a minimum, 100% line coverage (since it's Python).  Ensuring our test cases are rigorous is a separate issue.

PR #71, opened 2015-10-09
Improve Code Coverage

alexhersh:
In response to #67

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -1,4 +1,6 @@
 #!/bin/python
+# Import print_function to use a testable "print()" function 
+# instead of keyword "print".
 from __future__ import print_function
 
 import os
@@ -738,6 +740,7 @@ class NetworkPlugin(object):
         """
         Pull metadata for namespace and return it and a generated NS tag
         """
+        assert self.namespace
         ns_tag = self._escape_chars('%s=%s' % ('namespace', self.namespace))
         return ns_tag
 
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -1,23 +1,27 @@
 #!/bin/python
-import json
+from __future__ import print_function
+
 import os
-import sys
+import sh
 import re
+import sys
+import json
 import socket
+import logging
+import requests
+
 from docker import Client
 from docker.errors import APIError
-from subprocess import check_output, CalledProcessError, check_call
-import requests
-import sh
-import logging
-from netaddr import IPAddress, IPNetwork, AddrFormatError
 from logutils import configure_logger
+from netaddr import IPAddress, IPNetwork, AddrFormatError
+from subprocess import check_output, CalledProcessError, check_call
+
 import pycalico
 from pycalico import netns
-from pycalico.datastore import IF_PREFIX
-from pycalico.util import generate_cali_interface_name, get_host_ips
 from pycalico.ipam import IPAMClient
+from pycalico.datastore import IF_PREFIX
 from pycalico.block import AlreadyAssignedError
+from pycalico.util import generate_cali_interface_name, get_host_ips
 
 logger = logging.getLogger(__name__)
 pycalico_logger = logging.getLogger(pycalico.__name__)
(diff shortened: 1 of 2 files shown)


alexhersh:
Branching off of in-progress work by @Symmetric 
Currently, the only un-accounted for lines involve `calicoctl` calls that are subject to change soon.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -1,23 +1,27 @@
 #!/bin/python
-import json
+from __future__ import print_function
```
caseydavenport: Unused import?

alexhersh: nope its used once. print_function turns `print "statement"` into `print("statement")`.
The print_function can be unittested but __builtin__.print can't be easily tested

alexhersh: Importing this tells the compiler to treat `print` as a function and not a keyword. so `print "statement"` becomes `print("statement")`. This makes it easier to unittest.

caseydavenport: Right - I'd suggest adding a comment to explain that then, so someone doesn't come along and remove it without having that knowledge.

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -108,11 +116,104 @@ def test_delete(self):
...
-            self.assertEqual(namespace, self.plugin.namespace)
-            self.assertEqual(pod_name, self.plugin.pod_name)
-            self.assertEqual(docker_id, self.plugin.docker_id)
-            self.assertEqual(profile_name, self.plugin.profile_name)
-            m_datastore_client.remove_profile(profile_name)
+            assert_equal(namespace, self.plugin.namespace)
+            assert_equal(pod_name, self.plugin.pod_name)
+            assert_equal(docker_id, self.plugin.docker_id)
+            assert_equal(profile_name, self.plugin.profile_name)
+            self.m_datastore_client.remove_profile(profile_name)
```
Symmetric: Why are you calling this mock function?

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -108,11 +116,104 @@ def test_delete(self):
...
+            self.m_datastore_client.remove_profile.side_effect = KeyError
+
+            # Set up args
+            namespace = 'ns'
+            pod_name = 'pod1'
+            docker_id = 123456789101112
+            profile_name = 'ns_pod1_123456789101'
+
+            # Call method under test
+            self.plugin.delete(namespace, pod_name, docker_id)
```
Symmetric: Much neater -- but I think you should add a comment trailing this function saying something like "We don't need to assert anything here; if WAD we won't raise an exception."

It's a fine optimization to make (which is why I suggested it), but folk reading this code later will wonder if you just forgot to write asserts unless you call out why you omitted them.

calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -108,11 +116,104 @@ def test_delete(self):
...
+            assert_equal(pod_name, self.plugin.pod_name)
+            assert_equal(docker_id, self.plugin.docker_id)
+            assert_equal(profile_name, self.plugin.profile_name)
+            self.m_datastore_client.remove_profile(profile_name)
+
+    def test_delete_error(self):
+        """Test Pod Deletion Hook Failure"""
+        """
+        If the datastore remove_profile function returns KeyError, the profile is not in the datastore.
+        Issue warning log, but do not fail.
```
Symmetric: Can you rework this comment please? Characters are cheap, so let's have some full sentences. Something like "For this failure case we expect a warning log, but delete() should not raise an exception". (I found this sentence a bit hard to follow with my "knows nothing about the code" hat on).

Symmetric:
A few more points of review feedback; mostly minor markups, so I think next round should be the last.

calico_kubernetes/tests/kube_plugin_test.py
alexhersh: ooh thats a weird thing that was grandfathered in. I'll fix that.

Symmetric:
LGTM

Issue #66, opened 2015-10-06
Plugin logs do not wrap


caseydavenport:
The plugin generates log output to `/var/log/calico/kubernetes/calico.log` but this log doesn't have any log wrapping which can lead to giant logfiles.

We should follow the example of the other Calico component logs, and put plugin logs in their own directory.

For example,
`/var/log/calico/kubernetes/exec-plugin/current.log`

Symmetric:
The cheapo fix is to just use a [RotatingFileHandler](https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler) logger, rather than a FileHandler.

alexhersh:
Used the builtin rotating file handler instead of the `svlogd` process. Logs fill at 10M and 5 rotations are saved.

PR #73, opened 2015-10-09
Use Rotating File Handler for logs

alexhersh:
#66
Change the FileHandler to rotate when the file gets sufficiently large.

### DIFF
--- a/calico_kubernetes/logutils.py
+++ b/calico_kubernetes/logutils.py
@@ -3,7 +3,7 @@ import logging
 import os
 import sys
 
-LOG_DIR = '/var/log/calico/kubernetes'
+LOG_DIR = '/var/log/calico/kubernetes/'
 ROOT_LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(message)s'
 LOG_FORMAT = '%(asctime)s %(process)d %(levelname)s %(filename)s: %(message)s'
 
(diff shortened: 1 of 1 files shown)


alexhersh:
Alternatively we can use an `svlogd` process like the other Calico components, but that might get a bit messy with an exec plugin.

Symmetric:
Let's keep it simple; I don't see any reason to add another dependency when the Python builtin function works fine.

(svlogd is probably going away when we decompose our node container into single-process containers anyway).

calico_kubernetes/logutils.py
```diff
@@ -27,7 +27,9 @@ def configure_logger(logger, log_level, root_logger=False, log_dir=LOG_DIR):
     if not os.path.exists(log_dir):
         os.makedirs(log_dir)
 
-    file_hdlr = logging.FileHandler(filename=log_dir+'/calico.log')
+    file_hdlr = logging.handlers.RotatingFileHandler(filename=log_dir+'calico.log',
+                                                     maxBytes=25000,
```
Symmetric: 25K is very small - I'd go with something like 10M * 5 files.

calico_kubernetes/logutils.py
```diff
@@ -3,7 +3,7 @@
 import os
 import sys
 
-LOG_DIR = '/var/log/calico/kubernetes'
+LOG_DIR = '/var/log/calico/kubernetes/exec-plugin/'
```
Symmetric: Why have you added another directory? Does kubernetes/ have anything in it except this log? I'd be inclined to keep it as it is (or even use /var/log/calico/kubernetes-plugin.log instead of /kubernetes/calico.log)

alexhersh: This was based on a comment from Casey in the original Issue. I think he had a future process like the Policy Agent in mind.

Symmetric: I think that was assuming that we'd use svlogd. I'd prefer keeping it as-is for now; if we add a policy agent then we can refactor the logs then.

caseydavenport: Yep - if we're just using a RotatingFileHandler let's keep the same directory structure.

alexhersh:
@Symmetric, Good to go?

Symmetric:
LGTM

PR #74, opened 2015-10-15
Fix _log_interfaces function when ns is an int

Symmetric:
_log_interfaces didn't handle the case where the ns was an int.

Explicitly cast to a string in the function, for safety.


caseydavenport:
LGTM

PR #75, opened 2015-10-19
Skip `if __name__ == '__main__'` in coverage

Symmetric:
It's not easy to test the entry point to our plugin, so skip that
line in UTs. Everything will break if this isn't working, so
assume that FV/e2e tests will catch issues here.


caseydavenport:
LGTM

PR #77, opened 2015-10-20
Append Docker ID to logging messages

### LINKED ISSUE #72: Logs are impossible to follow in parallel execution
Kubelet runs pod updates concurrently, so we see logs from multiple plugin invocations interleaved. It's impossible to tell which log comes from which invocation.

We should prefix each log with a correlator, probably the 12-char concatenated docker ID.

We should be able to bake this into the log command and formatter so it doesn't need to be explicitly passed to every log statement.

alexhersh:
+ Fix #72
+ Use custom LoggerAdapter class to update the logger object with processed message after docker_id has been found.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Import print_function to use a testable "print()" function 
+# Import print_function to use a testable "print()" function
 # instead of keyword "print".
 from __future__ import print_function
 
@@ -26,7 +26,7 @@ import socket
 
 from docker import Client
 from docker.errors import APIError
-from logutils import configure_logger, configure_stdout_logger
+
 from netaddr import IPAddress, IPNetwork, AddrFormatError
 from policy import PolicyParser
 from subprocess import check_output, CalledProcessError, check_call
@@ -39,8 +39,10 @@ from pycalico.datastore_datatypes import Rule, Rules
 from pycalico.ipam import IPAMClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 
-logger = logging.getLogger(__name__)
+from logutils import *
+
 pycalico_logger = logging.getLogger(pycalico.__name__)
+logger = logging.getLogger(__name__)
 LOG_LEVEL = os.environ.get('LOG_LEVEL', 'INFO').upper()
 
 DOCKER_VERSION = "1.16"
@@ -726,15 +728,52 @@ def run_protected():
     """
     Runs the plugin, intercepting all exceptions.
     """
-    # Configure logging
-    configure_logger(logger, LOG_LEVEL, True)
-    configure_logger(pycalico_logger, LOG_LEVEL, False)
+    # Parse arguments and configure logging
+    global logger, pycalico_logger
+    mode = sys.argv[1]
+    namespace = sys.argv[2].replace('/', '_') if len(sys.argv) >=3 else None
+    pod_name = sys.argv[3].replace('/', '_') if len(sys.argv) >=4 else None
+    docker_id = sys.argv[4] if len(sys.argv) >=5 else None
+
+    # Append a stdout logging handler to log to the Kubelet.
+    # We cannot do this in the status hook because the Kubelet looks to
+    # stdout for status results.
+    log_to_stdout = (mode != 'status')
+
+    # Filter the logger to append the Docker ID to logs.
+    # If docker_id is not supplied, do not include it in logger config.
+    if docker_id:
+        configure_logger(logger=logger,
+                         log_level=LOG_LEVEL,
+                         log_format=DOCKER_ID_ROOT_LOG_FORMAT,
+                         log_to_stdout=log_to_stdout)
+        configure_logger(logger=pycalico_logger,
+                         log_level=LOG_LEVEL,
+                         log_format=DOCKER_ID_LOG_FORMAT,
+                         log_to_stdout=log_to_stdout)
+
+        docker_filter = IdentityFilter(identity=str(docker_id)[:12])
+        logger.addFilter(docker_filter)
+        pycalico_logger.addFilter(docker_filter)
+    else:
+        configure_logger(logger=logger,
+                         log_level=LOG_LEVEL,
+                         log_format=ROOT_LOG_FORMAT,
+                         log_to_stdout=log_to_stdout)
+        configure_logger(logger=pycalico_logger,
+                         log_level=LOG_LEVEL,
+                         log_format=LOG_FORMAT,
+                         log_to_stdout=log_to_stdout)
 
     # Try to run the plugin, logging out any BaseExceptions raised.
     logger.info("Begin Calico network plugin execution")
+    logger.info('Plugin Args: %s', sys.argv)
     rc = 0
     try:
-        run()
+        run(mode=mode,
+            namespace=namespace,
+            pod_name=pod_name,
+            docker_id=docker_id)
     except SystemExit:
         # If a SystemExit is thrown, we've already handled the error and have
         # called sys.exit().  No need to produce a duplicate exception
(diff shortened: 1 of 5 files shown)


calico_kubernetes/logutils.py
```diff
@@ -46,3 +46,10 @@ def configure_logger(logger, log_level, root_logger=False, log_dir=LOG_DIR):
     logger.addHandler(file_hdlr)
     logger.addHandler(stdout_hdlr)
     logger.setLevel(log_level)
+
+class DockerIDLogger(logging.LoggerAdapter):
+    """
+    Process Logger messages to append the Docker ID.
+    """
+    def process(self, msg, kwargs):
+        return '[%s] %s' % (self.extra['docker_id'], msg), kwargs
```
Symmetric: If we can avoid prepending the docker_id like this, I'd prefer it; the 'standard' log format is "TIMESTAMP OTHERSTUFF" and it's a bit of a pain to read and parse if it's otherwise.

I don't think you need this subclass; just use a LoggerAdapter, called with the arguments that you're currently using. The default implementation of process() is to add the contents of the `extra` dict to the `kwargs` from the log call.

Symmetric:
I think you can make this simpler, with a better log format -- should be a quick change.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -727,14 +729,37 @@ def run_protected():
...
+
+    # Append a stdout logging handler to log to the Kubelet.
+    # We cannot do this in the status hook because the Kubelet looks to
+    # stdout for status results.
+    if mode != 'status':
+        configure_stdout_logger(base_logger, logging.INFO, True)
+        configure_stdout_logger(pycalico_logger, logging.INFO, False)
+
+    # Adapt the logger to append the Docker ID to logs.
+    # Docker ID is not guarenteed in `init`, so keep the default "unknown" id.
```
caseydavenport: The Docker ID is guaranteed to not be present in `init`.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -39,8 +40,9 @@
 from pycalico.ipam import IPAMClient
 from pycalico.util import generate_cali_interface_name, get_host_ips
 
-logger = logging.getLogger(__name__)
 pycalico_logger = logging.getLogger(pycalico.__name__)
+base_logger = logging.getLogger(__name__)
+logger = logging.LoggerAdapter(base_logger, {'docker_id': "unknown"})
```
Symmetric: Rather than write "[unknown]" in all the logs that occur when we don't have an ID, can't we just change the logger's formatter to add the `docker_id` variable at the point that we learn the id (i.e. line 751)? That way we don't print useless log characters in the no-ID case.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -726,15 +728,55 @@ def run_protected():
...
-    configure_logger(pycalico_logger, LOG_LEVEL, False)
+    # Parse arguments and configure logging
+    global logger
+    mode = sys.argv[1]
+
+    # Append a stdout logging handler to log to the Kubelet.
+    # We cannot do this in the status hook because the Kubelet looks to
+    # stdout for status results.
+    if mode != 'status':
+        configure_stdout_logger(logger=logger,
```
Symmetric: Can we get round this by just making all our logs to stderr? I think this fix makes the logic here a bit tough to follow.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -726,15 +728,55 @@ def run_protected():
...
+                                log_level=logging.INFO,
+                                root_logger=True)
+        configure_stdout_logger(logger=pycalico_logger,
+                                log_level=logging.INFO,
+                                root_logger=False)
+
+    # Adapt the logger to append the Docker ID to logs.
+    # Docker ID is not supplied in `init`, so do not include the it in logs.
+    if mode != 'init':
+        namespace = sys.argv[2].replace('/', '_')
```
Symmetric: I'd rather you did all the arg parsing in one place, at the top, and then did `if docker_id:` here; keeps the related function closer together, and makes the intent of this if/else branch cleaner.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -726,15 +728,55 @@ def run_protected():
...
+
+        configure_logger(logger=logger,
+                         log_level=LOG_LEVEL,
+                         log_format=DOCKER_ID_ROOT_LOG_FORMAT)
+        configure_logger(logger=pycalico_logger,
+                         log_level=LOG_LEVEL,
+                         log_format=DOCKER_ID_LOG_FORMAT)
+
+        logger = logging.LoggerAdapter(logger, {'docker_id': str(docker_id)[:12]})
+        logger = logging.LoggerAdapter(pycalico_logger, {'docker_id': str(docker_id)[:12]})
```
Symmetric: I don't think you want to overwrite our __name__ logger with the pycalico logger

calico_kubernetes/calico_kubernetes.py
alexhersh: stdout logger logic was handled in another PR ([#80](https://github.com/projectcalico/calico-kubernetes/pull/80)) I am just moving the call around.

calico_kubernetes/calico_kubernetes.py
alexhersh: true!

calico_kubernetes/logutils.py
```diff
@@ -31,18 +33,14 @@ def configure_logger(logger, log_level, root_logger=False, log_dir=LOG_DIR):
                                                      maxBytes=10000000,
```
Symmetric: I've just spotted that this isn't safe, since we'll be writing to this log file from multiple processes. Maybe look at https://pypi.python.org/pypi/ConcurrentLogHandler/0.9.1?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -726,15 +728,55 @@ def run_protected():
...
+    else:
+        namespace = pod_name = docker_id = None
+        configure_logger(logger=logger,
```
Symmetric: I think this might add a second stdout and file handler to the logger, rather than replace the existing handlers. Can you check if this produces doubled logs?

calico_kubernetes/calico_kubernetes.py
Symmetric: There's a problem here -- this doesn't change the logger that pycalico is using, it just updates what our local pycalico_logger points to. I think you'll need to do something like `pycalico.logger = logging.LoggerAdapter` instead, to update the reference there. I don't think there's a reason to keep the local reference pycalico_logger; you can just pass pycalico.logger wherever you would use that.

calico_kubernetes/calico_kubernetes.py
alexhersh: no filehandler exists before this if/else branch. There might be a streamhandler, in which case the filehandler is added alongside the streamhandler, which isn't an issue.

calico_kubernetes/calico_kubernetes.py
alexhersh: pycalico doesn't have a unified logger. I don't think there is a way to safely adapt pycalico logs

Symmetric: Curses. Maybe we need to rethink this approach then. Instead of a LoggingAdapter, what about a Filter? It can cover similar use-cases. See https://docs.python.org/2/howto/logging-cookbook.html#using-filters-to-impart-contextual-information.

I think that might be clearer, because you can then do

```python
log_filter = IdentifierFilter(docker_id='xxx')
logger.addFilter(log_filter)
pycalico_logger.addFilter(log_filter)
```

Symmetric: Probably worth a quick prototype, then design chat with me or Casey; this issue has bounced around quite a lot, so I don't want to waste your time with more rewrites.

calico_kubernetes/calico_kubernetes.py
Symmetric: What about `if mode!=status and mode != init`? Then we've set up a stdout logger above, and will hit this branch too.

Symmetric:
One minor naming nit, then I think we're good.

Though the CI tests are failing; you'll need to get them green before this can be merged.

Symmetric:
LGTM!

Issue #79, opened 2015-10-22
Kubelet is not being informed of IP address updates


L-MA:
Logging is currently being streamed to stdout. This causes the Kubernetes status hook to fail since Kubernetes looks to stdout for status hook results. I.e.
```
Oct 20 00:30:43 kube-node1 kubelet[942]: E1020 00:30:43.193752     942 manager.go:343] NetworkPlugin calico failed on the status hook for pod 'no-access' - invalid character '-' after top-level value
```
Therefore, Kubernetes fails to learn about the IP address that Calico assigns to a pod.

caseydavenport:
Note this only applies when using CALICO_IPAM=true.  Fixed in v0.4.0

Symmetric:
Have you actually observed this to cause a failure? My reading of the code is that this error is benign (and in the past, when we didn't explicitly support the status hook, it was so):

```go
		if containerName == PodInfraContainerName {
			if inspectResult.NetworkSettings != nil {
				result.ip = inspectResult.NetworkSettings.IPAddress
			}
			// override the above if a network plugin exists
			if dm.networkPlugin.Name() != network.DefaultPluginName {
				netStatus, err := dm.networkPlugin.Status(pod.Namespace, pod.Name, kubeletTypes.DockerID(dockerID))
				if err != nil {
					glog.Errorf("NetworkPlugin %s failed on the status hook for pod '%s' - %v", dm.networkPlugin.Name(), pod.Name, err)
				} else if netStatus != nil {
					result.ip = netStatus.IP.String()
				}
			}
		}
```

caseydavenport:
Yes - this is an error when we are using CALICO_IPAM=true, since Calico assigns an IP address which is not the one assigned by Docker.

Kubernetes will assume we used the Docker IP address which means at least:
1) The Kubernetes API will report the wrong IP.
2) The kube-proxy will install incorrect proxying rules for services.

Symmetric:
Gotcha, thanks.

PR #80, opened 2015-10-22
Only configure stdout logging in setup and teardown hooks

L-MA:
Fix to issue #79 

Also includes bumping libcalico to the correct version that supports specifying rules at profile creation

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -26,7 +26,7 @@ import socket
 
 from docker import Client
 from docker.errors import APIError
-from logutils import configure_logger
+from logutils import configure_logger, configure_stdout_logger
 from netaddr import IPAddress, IPNetwork, AddrFormatError
 from policy import PolicyParser
 from subprocess import check_output, CalledProcessError, check_call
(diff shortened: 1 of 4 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -767,6 +767,12 @@ def run():
             logger.info('Executing Calico pod-status hook')
             NetworkPlugin().status(namespace, pod_name, docker_id)
         else:
+            # Append an stdout logging handler to log to the Kubelet
+            # We cannot do this in status hook because K8 looks to stdout
```
caseydavenport: Spell out Kubelet here rather than abbreviate as "K8"

caseydavenport:
One comment then LGTM.  Make sure it still has 100% coverage.

Symmetric:
@caseydavenport can we make this fix a 1-liner by changing our previous stdout logs to use stderr instead? I'm not crazy about conditionally changing where we make logs, and I suspect that Kubelet will capture stderr in the same way.

caseydavenport:
@Symmetric - Hmm, yeah, that could be a much better way to do this.  We should check that the `kubelet` behaves the way we expect in this case, but if it does, then let's make that change.

PR #82, opened 2015-10-23
Catch exceptions when logging interfaces

### LINKED ISSUE #81: netns PID missing causes logging to fail
As seen here: http://stackoverflow.com/questions/31265993/docker-networking-namespace-not-visible-in-ip-netns-list

The PID no longer exists after we delete the Docker created veth, and we attempt to use `ip netns exec`, which fails since it cannot find the symlink in `/var/run/netns`.

1) We shouldn't fail execution if debug logging fails
2) We shouldn't be relying on a symlink that may or may not exist.

caseydavenport:
fixes #81

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -158,7 +158,7 @@ class NetworkPlugin(object):
                                "to endpoint %s", self.docker_id)
             ip = ip_net[0].ip
 
-        logger.info("Retrieved IP Address: %s", ip)
+        logger.info("Retrieved pod IP Address: %s", ip)
 
         json_dict = {
             "apiVersion": "v1beta1",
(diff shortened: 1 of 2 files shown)


caseydavenport:
@Symmetric - Could you take a look at this? I whipped it up in response to an issue, but I think you'll be able to say whether or not it is an appropriate solution, or if I mis-judged the problem.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -709,18 +709,24 @@ def _log_interfaces(namespace):
...
+        if logger.isEnabledFor(logging.DEBUG):
+            interfaces = check_output(['ip', 'addr'])
+            logger.debug("Interfaces in default namespace:\n%s", interfaces)
+
+            namespaces = check_output(['ip', 'netns', 'list'])
+            logger.debug("Namespaces:\n%s", namespaces)
+
+            # Use nsenter because it doesn't rely on
+            # symlinks in /var/run/netns.
+            cmd = ['nsenter', '-t', str(namespace), '-n', 'ip', 'addr']
```
Symmetric: `nsenter` isn't installed on all systems. I'd suggest sticking with the `ip netns` command here, though you could make a check to see if `/var/run/netns/<ns>` exists first, and print a friendly warning in that case.

Definitely good to pokemon-catch here as well though; we shouldn't ever fail in this branch.

caseydavenport:
Good to go?

Symmetric:
LGTM

Symmetric:
...Though it looks like you'll need to rebase

PR #85, opened 2015-10-26
Add DNS config

caseydavenport:


Symmetric:
Which guides does this change affect?

caseydavenport:
This config will be compatible with both our bare-metal Ubuntu (systemd) guide and our bare-metal CoreOS guide for v1.1.

Systemd/Ubuntu5.04: https://github.com/kubernetes/kubernetes/pull/16204
CoreOS PR: Coming soon

PR #86, opened 2015-10-27
[WIP] Add cloud-config files for CoreOS demo

caseydavenport:
Depends on Kubernetes v1.1 release - don't merge until released.

Config used by: https://github.com/kubernetes/kubernetes/pull/16380

### DIFF
--- a/config/cloud-config/master-config-template.yaml
+++ b/config/cloud-config/master-config-template.yaml
@@ -1,0 +1,184 @@
+#cloud-config
+---
+hostname: kube-master
+users:
+  - name: core
+    ssh-authorized-keys: 
+      - <SSH_PUBLIC_KEY>
+    groups: 
+      - sudo
+    shell: /bin/bash
+
+write_files:
+  - path: /opt/bin/kubernetes-install.sh
+    owner: root
+    permissions: 0755
+    content: |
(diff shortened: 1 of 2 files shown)


caseydavenport:
@Symmetric - if you have a spare minute, could you take a look to make sure this all looks sensible before I merge?

It works in accordance with the guide, but I'd like a second pair of eyes.

Symmetric:
Looks fine. I'm sure we discussed this before, but I can't recall why we're not basing this off something like https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/coreos/coreos_multinode_cluster.md - I suspect we'd need https://github.com/projectcalico/calico-kubernetes/issues/93 in this case, since it's harder (though not impossible?) to manually edit the unit files under CoreOS.

Symmetric:
Also this punts on token auth -- which is something we'll want to include in all the docs I believe. (I'm just realizing this is going to be awkward under Vagrant, too)

caseydavenport:
@Symmetric - currently not basing off of that guide since it is quite a bit older.  This uses containerized Kubernetes processes rather than systemd, 

This also sets up a cluster that looks very similar to the one that our Ubuntu guide sets up.

PR #88, opened 2015-11-05
Change CALICO_IPAM default to "true"

### LINKED ISSUE #76: Enable CALICO_IPAM by default
Now that we have sufficient field testing, we should be enabling CALICO_IPAM by default.

This is a minor change to the code and an update to https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md

caseydavenport:
Fixes #76

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -55,10 +55,6 @@ ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 if ETCD_AUTHORITY_ENV not in os.environ:
     os.environ[ETCD_AUTHORITY_ENV] = 'kubernetes-master:6666'
 
-# Get the desired calicoctl location.  By default, we search the PATH for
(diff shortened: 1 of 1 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -68,7 +68,7 @@
 # Flag to indicate whether or not to use Calico IPAM.
 # If False, use the default docker container ip address to create container.
 # If True, use libcalico's auto_assign IPAM to create container.
-CALICO_IPAM = os.environ.get('CALICO_IPAM', 'false')
```
Symmetric: CALICO_IPAM=TRUE will fail. Should we force `os.environ.get().lower()` here?

(That's an existing bug, so happy to let this through if you don't want to fix as part of this PR)

caseydavenport: Happy to make that change here.

PR #89, opened 2015-11-05
Cleaned up info vs debug logs

caseydavenport:
Reduces a lot of logging when running at info level.  Debug can be enabled to get more detailed output.

Also, skips the status hook for host-networked pods since we don't configure networking for host-networked pods.


caseydavenport:
Closing in favor of #90, which includes these changes.

PR #90, opened 2015-11-06
Validation status fix + API access improvements

alexhersh:
+ Contains #89 
+ Host ns containers are not managed by Calico, but the kubelet will still use the plugin. This ensures that Calico does not fail setup/status calls. 
+ Improve api access in `get_pod_config`

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -101,16 +101,19 @@ class NetworkPlugin(object):
         self.profile_name = "%s_%s_%s" % (self.namespace,
                                           self.pod_name,
                                           str(self.docker_id)[:12])
-
-        logger.info('Configuring docker container %s', self.docker_id)
+        logger.info('Configuring pod %s/%s (container_id %s)', 
+                    self.namespace, self.pod_name, self.docker_id)
 
         try:
             endpoint = self._configure_interface()
+            logger.info("Created Calico endpoint: %s", endpoint.endpoint_id)
             self._configure_profile(endpoint)
         except CalledProcessError as e:
             logger.error('Error code %d creating pod networking: %s\n%s',
                          e.returncode, e.output, e)
             sys.exit(1)
+        logger.info("Successfully configured networking for pod %s/%s",
+                    self.namespace, self.pod_name)
 
     def delete(self, namespace, pod_name, docker_id):
         """Cleanup after a pod."""
@@ -121,23 +124,34 @@ class NetworkPlugin(object):
                                           self.pod_name,
                                           str(self.docker_id)[:12])
 
-        logger.info('Deleting container %s with profile %s',
-                    self.docker_id, self.profile_name)
+        logger.info('Removing networking from pod %s/%s (container id %s)',
+                    self.namespace, self.pod_name, self.docker_id)
 
         # Remove the profile for the workload.
         self._container_remove()
 
         # Delete profile
         try:
+            logger.info("Deleting Calico profile: %s", self.profile_name)
             self._datastore_client.remove_profile(self.profile_name)
         except:
             logger.warning("Cannot remove profile %s; Profile cannot "
                            "be found.", self.profile_name)
+        logger.info("Successfully removed networking for pod %s/%s",
+                    self.namespace, self.pod_name)
 
     def status(self, namespace, pod_name, docker_id):
         self.namespace = namespace
         self.pod_name = pod_name
         self.docker_id = docker_id
+
+        if self._uses_host_networking(self.docker_id):
+            # We don't perform networking / assign IP addresses for pods running
+            # in the host namespace, and so we can't return a status update 
+            # for them.
+            logger.debug("Ignoring status for pod %s/%s in host namespace",
+                         self.namespace, self.pod_name)
+            sys.exit(1)
 
         # Find the endpoint
         try:
@@ -147,24 +161,24 @@ class NetworkPlugin(object):
                 workload_id=self.docker_id
             )
         except KeyError:
-            logger.error("Container %s doesn't contain any endpoints",
-                         self.docker_id)
+            logger.error("Error in status: No endpoint for pod: %s/%s",
+                         self.namespace, self.pod_name)
             sys.exit(1)
 
         # Retrieve IPAddress from the attached IPNetworks on the endpoint
         # Since Kubernetes only supports ipv4, we'll only check for ipv4 nets
         if not endpoint.ipv4_nets:
-            logger.error("Exiting. No IPs attached to endpoint %s",
-                         self.docker_id)
+            logger.error("Error in status: No IPs attached to pod %s/%s",
+                         self.namespace, self.pod_name)
             sys.exit(1)
         else:
             ip_net = list(endpoint.ipv4_nets)
             if len(ip_net) is not 1:
                 logger.warning("There is more than one IPNetwork attached "
-                               "to endpoint %s", self.docker_id)
+                               "to pod %s/%s", self.namespace, self.pod_name)
             ip = ip_net[0].ip
 
-        logger.info("Retrieved pod IP Address: %s", ip)
+        logger.debug("Retrieved pod IP Address: %s", ip)
 
         json_dict = {
             "apiVersion": "v1beta1",
@@ -200,7 +214,7 @@ class NetworkPlugin(object):
         self._datastore_client.set_profiles_on_endpoint(
             [self.profile_name], endpoint_id=endpoint.endpoint_id
         )
-        logger.info('Finished configuring profile.')
+        logger.debug('Finished configuring profile.')
 
     def _configure_interface(self):
         """Configure the Calico interface for a pod.
@@ -229,7 +243,7 @@ class NetworkPlugin(object):
         interface_name = generate_cali_interface_name(IF_PREFIX,
                                                       ep.endpoint_id)
         node_ip = self._get_node_ip()
-        logger.info('Adding IP %s to interface %s', node_ip, interface_name)
+        logger.debug('Adding node IP %s to host-side veth %s', node_ip, interface_name)
 
         # This is slightly tricky. Since the kube-proxy sometimes
         # programs REDIRECT iptables rules, we MUST have an IP on the host end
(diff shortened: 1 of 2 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -778,11 +767,11 @@ def run_protected():
...
             pod_name=pod_name,
             docker_id=docker_id)
-    except SystemExit:
+    except SystemExit, e:
         # If a SystemExit is thrown, we've already handled the error and have
         # called sys.exit().  No need to produce a duplicate exception
-        # message, just set the return code to 1.
-        rc = 1
+        # message, just return the exit code.
+        rc = e
```
caseydavenport: I think you want `e.code` here.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -505,20 +504,9 @@ def _get_pod_ports(self, pod):
...
-            ns = pod['metadata']['namespace'].replace('/', '_')
-            name = pod['metadata']['name'].replace('/', '_')
-            if ns == self.namespace and name == self.pod_name:
-                this_pod = pod
-                break
-        else:
-            raise KeyError('Pod not found: ' + self.pod_name)
-        logger.debug('Got pod data %s', this_pod)
-        return this_pod
+        pod = self._get_api_path('namespaces/%s/pods/%s' % (self.namespace, self.pod_name))
```
caseydavenport: I like this change.

However, I think `_get_api_path` needs some thinking.  Currently, we only use it to get the `pod` API object.  However, its name suggests something more general purpose, and the docstring suggests it only returns a list of objects :)

I think keeping the name and updating the docstring to describe what is actually returned probably makes the most sense.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -141,6 +137,9 @@ def status(self, namespace, pod_name, docker_id):
                 workload_id=self.docker_id
             )
         except KeyError:
+            # Obtain information from Docker Client and validate container state
+            # Will exit safely (0) in host NetworkMode
+            self._validate_container_state(self.docker_id)
```
caseydavenport: I don't think this is the right abstraction.  We may or may not care about all of the things that currently, or might be one day be done in `_validate_container_state`.  We do know that we can't report a status for host-networked pods.

It is also a bit confusing to rely on the sys.exit in _validate_container_state here - makes the code harder to read.

I had actually already worked up a fix for this issue in another PR (which may or may not be the right way) - take a look here: https://github.com/projectcalico/calico-kubernetes/pull/89/files#diff-869f81c0955f1bd9cf59f97823a22585R142.

Let me know what you think, and we can discuss the right fix.

alexhersh: I like your solution, but we need to use my exit code handling so that the kubelet doesn't think we failed a status check.

calico_kubernetes/calico_kubernetes.py
alexhersh: We could probably just combine `_get_api_path` and `_get_pod_config`

calico_kubernetes/calico_kubernetes.py
caseydavenport: Sounds good.  How does the kubelet behave when it receives a exit code of 0, but no status is returned via stdout?  Is that the expected behavior?

calico_kubernetes/calico_kubernetes.py
caseydavenport: True, that.

calico_kubernetes/calico_kubernetes.py
alexhersh: from kubelet/network/exec/exec.go
```
//         - status, called at regular intervals and is supposed to return a json
//                formatted output indicating the pod's IPAddress(v4/v6). An empty string value or an erroneous output
//                will mean the container runtime (docker) will be asked for the PodIP
//                e.g. {
//                         "apiVersion" : "v1beta1",
//                         "kind" : "PodNetworkStatus",
//                         "ip" : "10.20.30.40"
//                     }
```

caseydavenport: Sounds good.

alexhersh: Rebased on your branch

calico_kubernetes/calico_kubernetes.py
```diff
@@ -504,44 +520,27 @@ def _get_pod_ports(self, pod):
...
+                     KUBE_API_ROOT)
         session = requests.Session()
-        if self.auth_token:
+        if self._api_root_secure() and self.auth_token:
             session.headers.update({'Authorization':
                                     'Bearer ' + self.auth_token})
-        response = session.get(KUBE_API_ROOT + path, verify=False)
-        response_body = response.text
+        try:
+            path = '%snamespaces/%s/pods/%s' % (KUBE_API_ROOT, self.namespace, self.pod_name)
```
caseydavenport: As a general rule, don't include anything in the try/except that doesn't need to be.  I'd move the `path = ... ` line outside of the try clause.

caseydavenport: Additionally, I think we need to be careful about how we're generating the path.

The way this is currently done, KUBE_API_ROOT _requires_ a trailing `/` - this could cause unexpected problems.  We should probably be resilient to both `http://api.url/api/v1` as well as `http://api.url/api/v1/`

caseydavenport: We also need to handle errors returned by `session.get(...)` - we should be logging out appropriate error messages when we hit an error.

`session.get(...)` will return a Response object, which may or may not have been successful.   We should check the return code and log out the response object (at debug level).  For bad return codes, we should log an error and exit gracefully.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -504,44 +520,27 @@ def _get_pod_ports(self, pod):
...
+        logger.debug('Getting API Resource: from KUBE_API_ROOT: %s',
+                     KUBE_API_ROOT)
         session = requests.Session()
-        if self.auth_token:
+        if self._api_root_secure() and self.auth_token:
             session.headers.update({'Authorization':
                                     'Bearer ' + self.auth_token})
-        response = session.get(KUBE_API_ROOT + path, verify=False)
-        response_body = response.text
+        try:
```
caseydavenport: Something like this is probably a good fit: https://docs.python.org/2/library/urlparse.html#urlparse.urljoin

alexhersh:
@caseydavenport care to take another look?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -500,44 +521,33 @@ def _get_pod_ports(self, pod):
...
-        _get_api_path('pods')
-
-        :param path: The relative path to an API endpoint.
-        :return: A list of JSON API objects
-        :rtype list
+        :return: JSON object containing the pod spec
         """
-        logger.info('Getting API Resource: %s from KUBE_API_ROOT: %s',
-                    path, KUBE_API_ROOT)
+        logger.debug('Getting API Resource: from KUBE_API_ROOT: %s',
```
caseydavenport: This log is useless and awkward now.  Rather than just remove the `%s` from it, we should either remove it or replace it with an appropriate log.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -500,44 +521,33 @@ def _get_pod_ports(self, pod):
...
-        return json.loads(response_body)['items']
+        # Ensure that the API_ROOT is a path to a directory,
+        # then join with the namespaced pod path.
+        root = KUBE_API_ROOT + ('/' if KUBE_API_ROOT[-1:] != '/' else '')
+        path = 'namespaces/%s/pods/%s' % (self.namespace, self.pod_name)
+        full_path = urljoin(root, path)
+        try:
+            response = session.get(full_path, verify=False)
+            logger.debug('Recieved API Response Object: %s', response)
+        except BaseException, e:
```
caseydavenport: This isn't quite right.

session.get(...) is not going to throw an error like you think it will.  If the request gets an error response (e.g. 404 not found, some other HTTP error), it is going to return successfully, but the `Response` object is going to have an error code.

We need to check that error code and act accordingly.

e.g (pseudo-code)
```
try:
  resp = session.get(url)
except SomeException:
  logger.exception("Exception querying API")
  sys.exit(1)
else:
    if resp.code != 200:
      log("Received error from API response: %s", resp)
      sys.exit(1)
```

calico_kubernetes/calico_kubernetes.py
```diff
@@ -500,44 +521,33 @@ def _get_pod_ports(self, pod):
...
                                     'Bearer ' + self.auth_token})
-        response = session.get(KUBE_API_ROOT + path, verify=False)
-        response_body = response.text
 
-        # The response body contains some metadata, and the pods themselves
-        # under the 'items' key.
-        return json.loads(response_body)['items']
+        # Ensure that the API_ROOT is a path to a directory,
+        # then join with the namespaced pod path.
+        root = KUBE_API_ROOT + ('/' if KUBE_API_ROOT[-1:] != '/' else '')
```
caseydavenport: Not a big fan of this approach.

I'd rather something like this (up at the top of the file).

```
KUBE_API_ROOT = env.get(...)
if not KUBE_API_ROOT.endswith("/"):
  KUBE_API_ROOT += "/"
```

Symmetric: I had some issues with `urljoin()` in another project; did you try using `os.path.join()`; I think it might be a simpler way of handling the possibly-missing trailing `/`.

alexhersh: I did experiment with that too, there was a gotcha that I can't seem to remember, but I'll look again. It may be fixed by verifying KUBE_API_ROOT

calico_kubernetes/calico_kubernetes.py
```diff
@@ -508,43 +528,35 @@ def _get_pod_ports(self, pod):
...
-            session.headers.update({'Authorization':
-                                    'Bearer ' + self.auth_token})
-        response = session.get(self.api_root + path, verify=False)
-        response_body = response.text
-
-        # The response body contains some metadata, and the pods themselves
-        # under the 'items' key.
-        return json.loads(response_body)['items']
+        with requests.Session() as session:
+            if self._api_root_secure() and self.auth_token:
```
caseydavenport: Could you stick a debug log here so we can trace execution better?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -508,43 +528,35 @@ def _get_pod_ports(self, pod):
...
+                session.headers.update({'Authorization':
+                                        'Bearer ' + self.auth_token})
+
+            path = os.path.join(self.api_root,
+                                'namespaces/%s/pods/%s' % (self.namespace,
+                                                           self.pod_name))
+            try:
+                logger.debug('Fetching API Resource from %s', path)
```
caseydavenport: "Fetching API resource from %s" is more generic that what is actually going on.

To be clear, this log should just say "Querying API for Pod: %s"

caseydavenport:
@alexhersh: A couple more comments regarding better log messages, but otherwise looks good.

Once those are fixed up I'll give a LGTM.

caseydavenport:
LGTM.

PR #94, opened 2015-11-13
Get configuration from on-disk file.

caseydavenport:
Configuration can now be performed through an on-disk config file.

The file lives in the same directory as the plugin, and takes the name "calico_kubernetes.ini".

The file has the following format:
```
[config]
ETCD_AUTHORITY=some-etcd-auth:4001
CALICO_IPAM=true
KUBE_API_ROOT=http://api:8080/api/v1/
...
```
For back-compatibility / flexibility, environment variables of the same names can still be used.  Environment variables take precedence over config file variables.

Not in this PR: configuration via `etcd`.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -54,13 +54,21 @@ CONFIG_FILENAME = "calico_kubernetes.ini"
 
 POLICY_ANNOTATION_KEY = "projectcalico.org/policy"
 
-# Values in configuration dictionary
+# Values in configuration dictionary.
 ETCD_AUTHORITY_VAR = "ETCD_AUTHORITY"
 LOG_LEVEL_VAR = "LOG_LEVEL"
 KUBE_AUTH_TOKEN_VAR = "KUBE_AUTH_TOKEN"
 KUBE_API_ROOT_VAR = "KUBE_API_ROOT"
 CALICO_IPAM_VAR = "CALICO_IPAM"
 DEFAULT_POLICY_VAR = "DEFAULT_POLICY"
+
+# All environment variables used by the plugin.
+ENVIRONMENT_VARS = [ETCD_AUTHORITY_VAR,
+                    LOG_LEVEL_VAR,
+                    KUBE_AUTH_TOKEN_VAR,
+                    KUBE_API_ROOT_VAR,
+                    CALICO_IPAM_VAR,
+                    DEFAULT_POLICY_VAR]
 
 
 class NetworkPlugin(object):
@@ -733,19 +741,9 @@ def load_config():
     # First, read the config file and get defaults.
     config = read_config_file()
 
-    # Get config from environment.
-    env_config = {
-            CALICO_IPAM_VAR: os.environ.get(CALICO_IPAM_VAR),
-            KUBE_API_ROOT_VAR: os.environ.get(KUBE_API_ROOT_VAR),
-            DEFAULT_POLICY_VAR: os.environ.get(DEFAULT_POLICY_VAR),
-            KUBE_AUTH_TOKEN_VAR: os.environ.get(KUBE_AUTH_TOKEN_VAR),
-    }
-
-    # Environment variables take precedence.
-    for k,v in env_config.iteritems():
-        if v is not None:
-            logger.debug("Use env variable: %s=%s", k, v)
-            config[k] = v
+    # Get config from environment, if defined.
+    for var in ENVIRONMENT_VARS:
+        config[var] = os.environ.get(var, config[var])
 
     # ETCD_AUTHORITY is handled slightly differently - we need to set it in the
     # environment so that libcalico works correctly.
@@ -754,6 +752,9 @@ def load_config():
                      ETCD_AUTHORITY_VAR,
                      config[ETCD_AUTHORITY_VAR])
         os.environ[ETCD_AUTHORITY_VAR] = config[ETCD_AUTHORITY_VAR]
+
+    # Ensure case is correct.
+    config[LOG_LEVEL_VAR] = config[LOG_LEVEL_VAR].upper()
 
     return config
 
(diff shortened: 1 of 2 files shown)


calico_kubernetes/tests/kube_plugin_test.py
```diff
@@ -894,32 +912,6 @@ def test_generate_rules_allow_all(self):
         # Assert return value is correct.
         assert_equal(return_val, expected)
 
-    def test_generate_rules_ns_iso(self):
```
caseydavenport: Duplicate test.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -724,6 +723,84 @@ def _log_interfaces(namespace):
...
+    Loads configuration for the plugin - returns a dictionary.
+
+    Looks first in environment, then in local config file.
+    """
+    # First, read the config file and get defaults.
+    config = read_config_file()
+
+    # Get config from environment.
+    env_config = {
+            CALICO_IPAM_VAR: os.environ.get(CALICO_IPAM_VAR),
```
Symmetric: Could save code with:

```
for var in [CALICO_IPAM_VAR, ...]:
    config[var] = os.environ.get(var, config[var])
```

? Could even pull the ENV_VARS list out as a constant if you want to keep this function tidy.

Or could even make this more generic with

```
for var in config:
    os.environ.get(var, config[var])
```

Not sure if I prefer this, since it allows you to load any env variable you want into the plugin...

The existing code works, but I find this sort of repeated block to be error-prone (easy to fail to update one of the calls down the road, for example).

calico_kubernetes/calico_kubernetes.py
```diff
@@ -724,6 +723,84 @@ def _log_interfaces(namespace):
...
+    """
+    Reads the config file on disk and returns configuration dictionary.
+    """
+    # Get the current directory and find path to config file.
+    executable = sys.argv[0]
+    cur_dir = os.path.dirname(executable)
+    config_file = os.path.join(cur_dir, CONFIG_FILENAME)
+
+    # Check that the file exists.
+    if not os.path.exists(config_file):
```
Symmetric: To be pedantic, I think you want `os.path.isfile()` here. Unlikely to matter in practice.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -724,6 +723,84 @@ def _log_interfaces(namespace):
...
+    # Read the config file.
+    parser = ConfigParser.ConfigParser(defaults)
+    parser.read(config_file)
+
+    # Make sure the config section exists
+    if not "config" in parser.sections():
+        sys.exit("No [config] section in file %s" % config_file)
+
+    # Get any values from the configuration file and populate dictionary.
+    config[ETCD_AUTHORITY_VAR] = parser.get("config", ETCD_AUTHORITY_VAR)
```
Symmetric: Again, can you streamline this with something like

```
for var in ENV_VARS:
    config[var] = parser.get("config", var)
```

?

Symmetric:
I like it. A couple suggestions for spots that you could streamline the code; up to you whether you prefer those modifications.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Yep, makes sense.  done.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Fixed.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Aye.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -724,6 +731,73 @@ def _log_interfaces(namespace):
...
+    """
+    Reads the config file on disk and returns configuration dictionary.
+    """
+    # Get the current directory and find path to config file.
+    executable = sys.argv[0]
+    cur_dir = os.path.dirname(executable)
+    config_file = os.path.join(cur_dir, CONFIG_FILENAME)
+
+    # Check that the file exists.
+    if not os.path.isfile(config_file):
```
caseydavenport: Wondering if this is the right behavior - you might still want to exclusively use environment variables, and this prevents that (you at least need a placeholder file).

caseydavenport:
Think I addressed it all.  One more question for you about behavior when no file is present.

calico_kubernetes/calico_kubernetes.py
```diff
@@ -724,6 +731,73 @@ def _log_interfaces(namespace):
...
+    parser = ConfigParser.ConfigParser(defaults)
+    parser.read(config_file)
+
+    # Make sure the config section exists
+    if not "config" in parser.sections():
+        sys.exit("No [config] section in file %s" % config_file)
+
+    # Get any values from the configuration file and populate dictionary.
+    for var in ENVIRONMENT_VARS:
+        config[var] = parser.get("config", var)
```
Symmetric: Oops -- that should be `get("config", config[var])`

calico_kubernetes/calico_kubernetes.py
Symmetric: Yeah, agree -- return the defaults if the file doesn't exist. Good spot.

calico_kubernetes/calico_kubernetes.py
caseydavenport: Nope, its a bit weird.

I pass in default values when I create the ConfigParser above.  This call is getting the value of `var` from the section `config`, so I think this is right.

caseydavenport:
Ok, all set I think.

calico_kubernetes/calico_kubernetes.py
Symmetric: Oh right -- gotcha.

Symmetric:
LGTM.

PR #95, opened 2015-11-13
Log to stderr instead of stdout.

caseydavenport:
Simplifies logging logic - still need to test this.


Symmetric:
Loving all the red; LGTM. I'll leave to you to merge once you've tested.

PR #96, opened 2015-11-13
Revamp cloud-config files for Conformance

alexhersh:
Use Secure API Authentication.
Update Manifests


caseydavenport:
Have you tested this?

It all looks sensible to me, and we should get it in as soon as it has been tested and shown to work.

alexhersh:
Not yet, I'm in the process of doing that

config/cloud-config/master-config-template.yaml
```diff
@@ -64,8 +66,8 @@ write_files:
         echo "Manifests not installed - installing."
 
         # Download config.
-        wget https://github.com/projectcalico/calico-kubernetes/archive/master.tar.gz
-        tar -xvf master.tar.gz
+        wget https://github.com/alexhersh/calico-kubernetes/archive/update-config-1.1.1.tar.gz
```
caseydavenport: Can you remove these changes?

config/node/network-environment-template
```diff
@@ -10,7 +10,7 @@ KUBERNETES_MASTER=<KUBERNETES_MASTER>
 ETCD_AUTHORITY=<KUBERNETES_MASTER>:6666
 
 # URL to access the Kubernetes apiserver
-KUBE_API_ROOT=http://<KUBERNETES_MASTER>:8080/api/v1/
+KUBE_API_ROOT=https://<KUBERNETES_MASTER>:443/api/v1/
```
caseydavenport: This is used for the ubuntu guide, which doesn't use HTTPs, so it should stay at 8080 for now.

config/cloud-config/master-config-template.yaml
```diff
@@ -20,17 +20,18 @@ write_files:
...
         tar -xf kubernetes.tar.gz
         tar -xf kubernetes/server/kubernetes-server-linux-amd64.tar.gz
-        
+
         # Install the `kubelet` and `kubectl` binaries.
         sudo chmod +x kubernetes/server/bin/kubelet
         sudo chmod +x kubernetes/server/bin/kubectl
         sudo cp -f kubernetes/server/bin/kubelet /opt/bin
         sudo cp -f kubernetes/server/bin/kubectl /opt/bin
+        sudo rm -rf /kubernetes
```
caseydavenport: Was this causing a problem?

config/node/calico-node.service
```diff
@@ -8,7 +8,7 @@ After=docker.service
 EnvironmentFile=/etc/network-environment
 User=root
 PermissionsStartOnly=true
-ExecStart=/usr/bin/calicoctl node --ip=${DEFAULT_IPV4} --kubernetes --kube-plugin-version=v0.4.0 --detach=false
+ExecStart=/usr/bin/calicoctl node --ip=${DEFAULT_IPV4} --kubernetes --kube-plugin-version=v0.6.0 --detach=false
```
caseydavenport: v0.6.1 is the latest.

config/cloud-config/master-config-template.yaml
alexhersh: no, but its a huge directory and we need 2 files from it. I did the same for our calico-kubernetes clone too.

caseydavenport: Could you make that:
```
sudo rm -r ./kubernetes 
```

instead?

config/node/network-environment-template
alexhersh: These changes will go in as both the Coreos and Ubuntu guides are updated.

caseydavenport:
@alexhersh - Can you split this in to two PRs? One to update to Kubernetes v1.1.1, another to add in all the authentication?

The v1.1.1 changes don't need docs updates, but the other do, and I'd like to get the v1.1.1 change in sooner rather than later.

config/cloud-config/node-config-template.yaml
```diff
@@ -44,7 +47,7 @@ write_files:
       echo " " >> /etc/network-environment
```
caseydavenport: We should get rid of the whole /etc/network-environment thing for configuring Calico and use the config file method instead - it will simplify this a bunch.

alexhersh:
See https://github.com/alexhersh/kubernetes/pull/3 and https://github.com/alexhersh/kubernetes/pull/4 for doc updates

caseydavenport:
Closing in favor of the [calico-cni](https://github.com/projectcalico/libcalico/pull/77) version.

PR #97, opened 2015-11-13
Option to disable all policy

caseydavenport:
Adds the ability to specify `DEFAULT_POLICY=none`, thus disabling all policy in the plugin.

Annotations will not be used to generate policy, and no namespace isolation will be enforced.

This mode does not require the `apiserver` to be running, and does not query the API at all.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -52,7 +52,11 @@ HOSTNAME = socket.gethostname()
 # Config filename.
 CONFIG_FILENAME = "calico_kubernetes.ini"
 
+# Key to use when searching for annotations.
 POLICY_ANNOTATION_KEY = "projectcalico.org/policy"
+
+# Name of the profile to apply when using DEFAULT_POLICY == none.
+DEFAULT_PROFILE_NAME = "default-profile"
 
 # Values in configuration dictionary.
 ETCD_AUTHORITY_VAR = "ETCD_AUTHORITY"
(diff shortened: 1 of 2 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -192,29 +229,63 @@ def status(self, namespace, pod_name, docker_id):
...
-        logger.debug('Finished configuring profile.')
+            # Policy is disabled - add this pod to the default profile.
+            if not self._datastore_client.profile_exists(DEFAULT_PROFILE_NAME):
+                # If the default profile doesn't exist, create it.
+                logger.info("Creating profile '%s'", DEFAULT_PROFILE_NAME)
+                allow = Rule(action="allow")
+                rules = Rules(id=DEFAULT_PROFILE_NAME, 
+                              inbound_rules=[allow], 
+                              outbound_rules=[allow])
+                self._datastore_client.create_profile(DEFAULT_PROFILE_NAME,
```
Symmetric: Paranoid case: we check for the profile, it doesn't exist, so we try to create it. In the meantime, another process creates it. Will this call fail and take the whole function call down? Could we be more defensive, and re-check for existence if we fail here? If we fail and then the profile is there, we can continue.

Symmetric:
One error case to consider, which might not actually be hittable. LGTM otherwise, feel free to merge when that's addressed.

calico_kubernetes/calico_kubernetes.py
caseydavenport: I believe create_profile is idempotent, so this call shouldn't fail.

PR #100, opened 2015-11-17
Move logging filter to handlers

L-MA:
Filters, unlike levels and handlers, do not propogate to child loggers. Therefore, the child loggers of pycalico did not inherit the IdentityFilter. Fixed by attaching the filter to the handlers.

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -834,16 +834,15 @@ def run_protected():
     if docker_id:
         configure_logger(logger=logger,
                          log_level=config[LOG_LEVEL_VAR],
+                         docker_id=str(docker_id)[:12],
                          log_format=DOCKER_ID_ROOT_LOG_FORMAT,
                          log_to_stdout=log_to_stdout)
         configure_logger(logger=pycalico_logger,
                          log_level=config[LOG_LEVEL_VAR],
+                         docker_id=str(docker_id)[:12],
                          log_format=DOCKER_ID_LOG_FORMAT,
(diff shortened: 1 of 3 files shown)


calico_kubernetes/logutils.py
```diff
@@ -33,11 +33,14 @@ def configure_logger(logger, log_level, log_format=LOG_FORMAT,
         os.makedirs(log_dir)
 
     formatter = logging.Formatter(log_format)
+    docker_filter = IdentityFilter(identity=filter_id)
```
caseydavenport: This is awkward, since we're calling it docker_filter, but the argument is "identity" which is generic and says nothing about the docker_id.

We should probably just be specific and call it "docker_id" so it is clear what we mean - we're never passing anything else in, right?

calico_kubernetes/calico_kubernetes.py
```diff
@@ -834,23 +834,24 @@ def run_protected():
...
                          log_format=DOCKER_ID_LOG_FORMAT,
                          log_to_stdout=log_to_stdout)
 
-        docker_filter = IdentityFilter(identity=str(docker_id)[:12])
-        logger.addFilter(docker_filter)
-        pycalico_logger.addFilter(docker_filter)
     else:
         configure_logger(logger=logger,
                          log_level=config[LOG_LEVEL_VAR],
+                         docker_id=None,
```
Symmetric: Perhaps make this a kwarg with default `None`, then you don't need to specify this arg here?

Symmetric:
One minor code tidy suggestion, I'm happy for you to merge once you make that change.

PR #103, opened 2015-11-18
Remove validation done in libcalico

L-MA:
Wait for https://github.com/projectcalico/libcalico/pull/53 to get it

### DIFF
--- a/calico_kubernetes/policy.py
+++ b/calico_kubernetes/policy.py
@@ -59,7 +59,7 @@ class PolicyParser(object):
         try:
             args = docopt.docopt(__doc__, argv=splits)
         except docopt.DocoptExit:
-            raise PolicyParseError("Failed to parse policy: %s", policy)
+            raise ValueError("Failed to parse policy: %s", policy)
 
         # Generate a rule object from the arguments.
         rule = self._generate_rule(args)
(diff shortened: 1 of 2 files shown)


calico_kubernetes/policy.py
```diff
@@ -98,34 +98,24 @@ def _generate_rule(self, arguments):
         if protocol:
             rule_args["protocol"] = protocol
         if src_ports:
-            rule_args["src_ports"] = self._validate_ports(src_ports)
+            rule_args["src_ports"] = src_ports
```
caseydavenport: I think we still need something here to convert to a list, correct?

`src_ports` is a string which looks like this: `"10,20,30"` but it needs to looks like this: `[10, 20, 30]`

calico_kubernetes/policy.py
```diff
@@ -98,33 +98,36 @@ def _generate_rule(self, arguments):
...
+    def _sanitize_ports(self, value):
         """
-        Validates the given value can be parsed as an int.
-        TODO: Move validation code into libcalico
+        Splits the given string of ports, delimited by a comma, into a list.
+        Removes all whitespace before splitting.
+        Validation is done in libcalico.
+
+        :param value: Port string.  E.g. "80,8001"
+        :return: List of ports.  E.g [80, 8001]
```
caseydavenport: List of ports is actually a list of strings, so represent it like ["80", "8001"]

calico_kubernetes/policy.py
```diff
@@ -98,33 +98,36 @@ def _generate_rule(self, arguments):
...
         """
-        try:
-            return int(value)
-        except ValueError:
-            raise PolicyParseError("Failed to parse %s, expecting int", value)
+        value = value.replace(' ', '')
+        ports = value.split(",")
```
caseydavenport: Something like this would be more Pythonic:
```
ports = [s.strip() for s in value.split(",")]
```

caseydavenport: You could probably get rid of this whole function and inline this up above -

```
rule_args["src_ports"] = [s.strip() for s in value.split(",")]
```

calico_kubernetes/policy.py
```diff
@@ -98,33 +98,36 @@ def _generate_rule(self, arguments):
...
+        Removes all whitespace before splitting.
+        Validation is done in libcalico.
+
+        :param value: Port string.  E.g. "80,8001"
+        :return: List of ports.  E.g [80, 8001]
         """
-        try:
-            return int(value)
-        except ValueError:
-            raise PolicyParseError("Failed to parse %s, expecting int", value)
```
caseydavenport: I'm wondering about the PolicyParseError now.  `libcalico` throws ValueErrors, so I wonder if we should be doing the same to be consistent.

At a minimum, PolicyParseError should inherit from ValueError.

calico_kubernetes/tests/policy_test.py
```diff
@@ -135,18 +135,15 @@ def test_parse_icmp_rule(self):
         # Assert on the result
         assert_equal(parsed['action'], "allow")
         assert_equal(parsed['protocol'], "icmp")
-        assert_equal(parsed['icmp_type'], 8)
+        assert_equal(parsed['icmp_type'], '8')
         assert_equal(parsed['src_net'], IPNetwork("1.2.0.0/16"))
 
     @parameterized.expand([
         ("allow icmp type code 4"),
-        ("allow icmp type ten code 4"),
```
caseydavenport: why remove these test cases?

L-MA: calico_kubernetes uses a specific version of libcalico, not master, therefore, the libcalico validation changes are not reflected in these tests. Should we remove the version locking of libcalico or should we remove these test cases?

caseydavenport: Yep, we should probably be stuck on master libcalico.

caseydavenport:
@L-MA - when you've got a chance, can you bring this guy up to date?

calico_kubernetes/policy.py
```diff
@@ -98,73 +98,35 @@ def _generate_rule(self, arguments):
...
-
     def _validate_label(self, value):
         """
         Takes the given label, validates it, and returns the equivalent tag.
         """
         match = LABEL_REG.search(value)
         if not match:
-            raise PolicyParseError("Failed to parse %s, expecting "
+            raise ValueError("Failed to parse %s, expecting "
                                    "label of form X=Y", value)
```
caseydavenport: Awkward indentation here

caseydavenport:
One more comment to address then LGTM

PR #110, opened 2015-12-02
Simplify build

tomdee:


Makefile
```diff
@@ -1,36 +1,26 @@
...
 	
 	# Build the kubernetes plugin
 	docker run \
 	-u user \
-	-v `pwd`/calico_kubernetes:/code/calico_kubernetes \
-	-v `pwd`/dist:/code/dist \
-	-e PYTHONPATH=/code/calico_kubernetes \
-	calico/kubernetes-build pyinstaller calico_kubernetes/calico_kubernetes.py -a -F -s --clean
+	-v `pwd`:/code \
+	calico/build pyinstaller calico_kubernetes/calico_kubernetes.py -a -F -s --clean
```
caseydavenport: What is the recommended way to add new requirements if we're using a centralized build container?

caseydavenport:
Good amount of red in this PR.  Makes me feel good.

One concern about sharing build containers, but maybe I shouldn't be.

Makefile
tomdee: I think as long as the requirements aren't huge we just add them to the source container. I can go ahead and do that.

caseydavenport: Ok.  I'm a little worried that this could cause pain down the road, but it shouldn't be too much pain, so I'm happy to let it go for now and see how it works.

caseydavenport:
LGTM.

PR #115, opened 2015-12-03
Validate container state at beginning of create() hook

### LINKED ISSUE #114: Check for host networking before making changes
The plugin ignores host networked containers (e.g. kube-proxy).  However, due to a bug in our code, it doesn't check if the container is host-networked until AFTER it attempts to do some stuff.

It should check the NetworkMode first thing, and exit successfully without doing anything if it encounters a NetworkMode=host container.

caseydavenport:
Fixes #114


tomdee:
LGTM

PR #116, opened 2015-12-03
Fix previous build fixes

tomdee:
- Add coverage
- deal with user not be UID 1000
- Use the fact that the calico/test container now has the requried modules

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,15 @@ dist/calico_kubernetes: $(SRCFILES)
 	# Build the kubernetes plugin
 	docker run \
 	-u user \
-	-v `pwd`:/code \
+	-v `pwd`/dist:/code/dist \
+	-v `pwd`/calico_kubernetes:/code/calico_kubernetes \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,14 +14,15 @@ dist/calico_kubernetes: $(SRCFILES)
...
+	-v `pwd`/dist:/code/dist \
+	-v `pwd`/calico_kubernetes:/code/calico_kubernetes \
 	calico/build pyinstaller calico_kubernetes/calico_kubernetes.py -a -F -s --clean
 
 ut:
-	docker run --rm -v `pwd`:/code \
-	calico/test sh -c \
-	'pip install ConcurrentLogHandler && \
-	nosetests calico_kubernetes/tests -c nose.cfg'
+	docker run --rm  -v `pwd`/calico_kubernetes:/code/calico_kubernetes \
```
caseydavenport: You've got an extra space between `--rm` and `-v`

PR #118, opened 2015-12-04
Add "make" targets for create a Calico enabled kubernetes cluster

tomdee:
- Just run "make run-kubernetes-master" to get a single node Kubernetes cluster
- Also rename the binary from calico_kubernetes to just calico

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,26 @@
-.PHONY: all binary ut clean
+.PHONY: all binary ut clean run-service-proxy run-kubernetes-master
 
 SRCFILES=$(shell find calico_kubernetes)
+LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
+K8S_VERSION=1.1.2
 
 default: all
 all: binary test
 test: ut
-binary: dist/calico_kubernetes
+binary: dist/calico
 
-dist/calico_kubernetes: $(SRCFILES)
+dist/calico: $(SRCFILES)
 	mkdir -p dist
 	chmod 777 `pwd`/dist
-	
+
+	# Stop the master kubelet since if it's running it holds a lock on the file
+	-docker stop calico-kubelet-master
 	# Build the kubernetes plugin
 	docker run \
 	-u user \
 	-v `pwd`/dist:/code/dist \
 	-v `pwd`/calico_kubernetes:/code/calico_kubernetes \
-	calico/build pyinstaller calico_kubernetes/calico_kubernetes.py -a -F -s --clean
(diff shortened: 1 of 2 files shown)


caseydavenport:
This seems like a nice first step.  Going to take a deeper look.

The hardest part will, of course, getting this to a point where we can use it as a suitable ST rig.

Makefile
```diff
@@ -37,8 +41,59 @@ ut-circle: binary
 	--with-xunit --xunit-file=/circle_output/output.xml; RC=$$?;\
 	[[ ! -z "$$COVERALLS_REPO_TOKEN" ]] && coveralls || true; exit $$RC'
 
+dist/calicoctl:
+	mkdir -p dist
+	curl -L http://www.projectcalico.org/latest/calicoctl -o dist/calicoctl
+	chmod +x dist/calicoctl
+
+run-kubernetes-master: run-etcd run-service-proxy binary dist/calicoctl kubectl
```
caseydavenport: Do we need a `stop-kubernetes-master` make target to tear down the cluster?

Makefile
```diff
@@ -37,8 +41,59 @@ ut-circle: binary
...
+	-e KUBE_API_ROOT=http://localhost:8080/api/v1/ \
+	-d \
+	gcr.io/google_containers/hyperkube:v$(K8S_VERSION) \
+	/hyperkube kubelet --network-plugin calico --v=5 --containerized --hostname-override="127.0.0.1" --address="0.0.0.0" --api-servers=http://localhost:8080 --config=/etc/kubernetes/manifests
+
+	# Start the calico node
+	sudo dist/calicoctl node
+
+run-service-proxy:
+	-docker rm -f calico-service-proxy
```
caseydavenport: While it acts as a service proxy, it is always referred to as the "kube-proxy", so I'd keep that naming to avoid confusion.

Makefile
```diff
@@ -37,8 +41,59 @@ ut-circle: binary
...
+
+run-kubernetes-master: run-etcd run-service-proxy binary dist/calicoctl kubectl
+	# Stop any existing kubelet that we started
+	-docker rm -f calico-kubelet-master
+
+	# Remove any pods that the old kubelet may have started.
+	-docker rm -f $$(docker ps | grep k8s_ | awk '{print $$1}')
+
+	# Run the kubelet which will launch the master components in a pod.
+	docker run \
```
caseydavenport: If we're running the kubelet on the host docker, then the pods that it creates are going to start up using the host docker as well.  This won't work for the multi-node case.

I think we want the kubelet / master services all running on docker in docker, right?

tomdee: Yes, I'll make this change when we add supoprt for multi host

PR #119, opened 2015-12-04
Cleanup when we fail to setup neworking

### LINKED ISSUE #117: Plugin should clean up endpoints / IP addresses better
The plugin makes an attempt to clean up IP addresses and endpoints if it encounters an error, but it doesn't handle all cases.  The plugin should be more robust to cases like:

1) Creates endpoint.
2) Assigns IP.
3) Fails to configure profile.

Any time the plugin returns a non-zero exit code during a create, it needs to delete the endpoint and un-assign any IP addresses.

caseydavenport:
WIP - needs test.

Fixes #117

### DIFF
--- a/calico_kubernetes/calico_kubernetes.py
+++ b/calico_kubernetes/calico_kubernetes.py
@@ -424,11 +424,19 @@ class NetworkPlugin(object):
             try:
                 ipv4s, ipv6s = self._datastore_client.auto_assign_ips(1, 0,
                                                         self.docker_id, None)
-                ip = ipv4s[0]
                 logger.debug("IPAM assigned ipv4=%s; ipv6= %s", ipv4s, ipv6s)
             except RuntimeError as err:
-                logger.error("Cannot auto assign IPAddress: %s", err.message)
+                logger.error("Cannot auto assign IP address: %s", err.message)
                 sys.exit(1)
+
+            # Check to make sure an address was assigned.
+            if not ipv4s:
+                logger.error("Unable to assign an IP address - exiting")
+                sys.exit(1)
+
+            # Get the address.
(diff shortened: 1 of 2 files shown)


calico_kubernetes/calico_kubernetes.py
```diff
@@ -118,12 +118,24 @@ def create(self, namespace, pod_name, docker_id):
...
+            logger.exception("Error networking pod - cleaning up")
+
+            try:
+                self.delete(namespace, pod_name, docker_id)
+            except BaseException:
+                # Catch all errors tearing down the pod - this
+                # is best-effort.
+                logger.exception("Error cleaning up pod")
+
+            # We've torn down, now re-raise the Exception.
```
tomdee: reraise?

caseydavenport: Bugged comment, we just sys.exit().

calico_kubernetes/calico_kubernetes.py
```diff
@@ -118,12 +118,24 @@ def create(self, namespace, pod_name, docker_id):
...
-        except CalledProcessError as e:
-            logger.error('Error code %d creating pod networking: %s\n%s',
-                         e.returncode, e.output, e)
+        except BaseException:
+            # Check to see if an endpoint has been created.  If so,
+            # we need to tear down any state we may have created.
+            logger.exception("Error networking pod - cleaning up")
+
+            try:
+                self.delete(namespace, pod_name, docker_id)
```
tomdee: Does .delete() make a best effort to reverse all of the side effects of everything in the try: block? From a quick code read it looks like it doesn't. If that's OK, I think you need to add a comment to justify why.

PR #124, opened 2015-12-17
refactor calicoctl installer

PierreKircher:
https://asciinema.org/a/6vcysvgbwgbkwgb85lqjczrth

up/downgrade calicoctl + node container every 15 min without terminating the node 

falls back to 0.12.0 

simply etcdctl set /calico/version "0.10.0" and wait 15 min or reboot the node to trigger the up/downgrade

### DIFF
--- a/config/cloud-config/master-config-template.yaml
+++ b/config/cloud-config/master-config-template.yaml
@@ -38,19 +38,28 @@ write_files:
     permissions: 0755
     content: |
       #! /usr/bin/bash
-      
+      CALICO_VERSION=$(/usr/bin/etcdctl get /calico-install/version) >/dev/null 2>&1 || CALICO_VERSION="0.12.0"
       if [ ! -f /opt/bin/calicoctl ]; then
         echo "Calico not installed - installing."
-       
         # Install the `calicoctl` binary
-        wget https://github.com/projectcalico/calico-docker/releases/download/v0.12.0/calicoctl
+        wget https://github.com/projectcalico/calico-containers/releases/download/v${CALICO_VERSION}/calicoctl
         chmod +x calicoctl
         sudo mv calicoctl /opt/bin
-
         # Fetch the calico/node container
-        sudo docker pull calico/node:v0.12.0
+        sudo docker pull calico/node:v${CALICO_VERSION} 
+      else
+        INSTALLED_VERSION=$(/opt/bin/calicoctl version)
+        MATCH=$(echo "${INSTALLED_VERSION}" | grep -c "${CALICO_VERSION}")
+        if [ $MATCH -eq 0 ]; then
+          wget https://github.com/projectcalico/calico-containers/releases/download/v${CALICO_VERSION}/calicoctl
+          chmod +x calicoctl
+          sudo mv calicoctl /opt/bin
+          # Fetch the calico/node container
+          sudo docker pull calico/node:v${CALICO_VERSION} 
+        else
+          echo "OK! you are already uptodate"
+        fi
       fi
-
   - path: /opt/bin/install-manifests.sh
     owner: root
     permissions: 0755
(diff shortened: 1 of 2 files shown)


caseydavenport:
@PierreKircher - Thanks for this, looks like a nice feature.

Before we can accept any PRs, you'll need to sign our contributors agreement (CLA).  You can find out more about that here:  http://docs.projectcalico.org/en/latest/contribute.html

PierreKircher:
i signed that about an hour ago :)

config/cloud-config/master-config-template.yaml
```diff
@@ -38,19 +38,28 @@ write_files:
...
         sudo mv calicoctl /opt/bin
-
         # Fetch the calico/node container
-        sudo docker pull calico/node:v0.12.0
+        sudo docker pull calico/node:v${CALICO_VERSION} 
+      else
+        INSTALLED_VERSION=$(/opt/bin/calicoctl version)
+        MATCH=$(echo "${INSTALLED_VERSION}" | grep -c "${CALICO_VERSION}")
+        if [ $MATCH -eq 0 ]; then
+          wget https://github.com/projectcalico/calico-docker/releases/download/v${CALICO_VERSION}/calicoctl
```
caseydavenport: `calico-docker` was recently renamed to `calico-containers`, so worth updating this link to match.

config/cloud-config/master-config-template.yaml
```diff
@@ -38,19 +38,28 @@ write_files:
...
+          chmod +x calicoctl
+          sudo mv calicoctl /opt/bin
+          # Fetch the calico/node container
+          sudo docker pull calico/node:v${CALICO_VERSION} 
+        else
+          echo "OK! you are already uptodate"
```
caseydavenport: uptodate -> up-to-date

config/cloud-config/master-config-template.yaml
```diff
@@ -38,19 +38,28 @@ write_files:
     permissions: 0755
     content: |
       #! /usr/bin/bash
-      
+      CALICO_VERSION=$(/usr/bin/etcdctl get /calico/version) >/dev/null 2>&1 || CALICO_VERSION="0.12.0"
```
caseydavenport: I'm worried about namespacing this under `/calico` since this isn't part of the Calico etcd API and data-model, rather it is owned and used by these scripts (i.e this isn't the source of truth as for which version of Calico is installed).  I think this key should probably live elsewhere - perhaps `/calico-install/version` or something similar.

caseydavenport:
@PierreKircher - Sorry for the long delay on this guy.  Just a few minor comments from me.

PierreKircher:
@caseydavenport done .. have a look and let me know > if something is still missing hit me up on slack .. thanks

config/cloud-config/node-config-template.yaml
```diff
@@ -73,19 +73,28 @@ write_files:
     permissions: 0755
     content: |
       #! /usr/bin/bash
-      
+      CALICO_VERSION=$(/usr/bin/etcdctl get /calico-install/version) >/dev/null 2>&1 || CALICO_VERSION="0.12.0"
```
caseydavenport: Will `etcdctl get` work like this on the node? For this guide, `etcd` is running on the master.  I'd expect that we need to pass `--endpoint` (or set the env variable) as described here: https://github.com/coreos/etcd/tree/master/etcdctl#endpoint

PierreKircher: nope normaly thats done via env variables and it usualy runs even on a worker just in proxy mode so localhost is tryed first .. just the way etcd works :)

caseydavenport: Ah, so the idea is that each node will use its own local `etcd` instance to determine which version to run (so you can mix and match Calico versions) rather than each node looking to the master `etcd` to determine which Calico version to use?

caseydavenport:
Closing this since this repo is no longer in use.

